PR #10, opened 2015-05-29
avoid errors when no configuration is set

gelbehexe:
Pages can be shown even if there is no real url configuration


dmitryd:
I need to check this further because RealURL is expected to work with default values if there is no configuration. This should not be needed.

dmitryd:
I believe this is not needed because realurl provides default values for its operations.

Please, provide examples when it does not work without configuration.

dmitryd:
No comment provided. Closing the issue.

PR #12, opened 2015-09-04
[TASK] Added simple flow for TYPO3 7.4 or higher

bmoex:
See separate commits for more information


dmitryd:
Thanks! I only removed die() in ext_localconf.php because I do not see much sense in having it.

PR #13, opened 2015-09-05
Add "exclude page path" function

Razrael:
I activated the "Exclude from speaking URL" again and implemented it into encoder and decoder.

Since the excluded pages are not the default behavior, it will only search for them when nothing else was found. This should save some time on most page loads.


dmitryd:
@Razrael There is a problem here (see the screenshot). When the segment with this checkbox is lined, it should receive the full url (like it has no flag set). Right now its link is the same as the link of the parent page.
missing-segment

dmitryd:
Doing further work here: https://github.com/dmitryd/typo3-realurl/tree/feature/excludePage

PR #15, opened 2015-11-11
apiwrappers Compatibility for Typo3 6.1

guthypeter:
Hi,

we were able to use the extension with Typo3 6.1.7 only with this patch, as \TYPO3\CMS\Core\Utility\ArrayUtility::mergeRecursiveWithOverrule isn't defined here.


gkuehnberger:
Ahoi Dmitry, please include this patch in the next release, so that the advertized compliance with 6.* is also available for 6.1.
Thanks G

PR #20, opened 2015-12-21
[TASK] Updates the backend module for TYPO3 7.

echavaillaz:


dmitryd:
There is a problem: it will break TYPO3 4.5. There are many sites that still use RealURL there.

I suppose the fix will stay here for anybody, who has the issue, to apply manually. Meanwhile realurl v2 will be released soon with new BE module and will not have such issue.

echavaillaz:
There is nothing who will break TYPO3 4.5. I make the check for all changes.

dmitryd:
Namespaces in the code?

echavaillaz:
Ok

dmitryd:
Btw, you can try version 2. It now has Backend modules and it is feature complete.

dmitryd:
Incompatible with 4.5. Please, use realurl 2 instead.

PR #27, opened 2016-02-10
[TASK] Respects appendMissingSlash option

spoonerWeb:
Due there was no possibility to remove the trailing slash, I used the existing config "appendMissingSlash" (which I set to FALSE) to prevent the addition of the slash.


dmitryd:
Thank you!

PR #30, opened 2016-02-22
[BUGFIX] Use parent id upon overlay match in decoding process

fsuter:
When decoding the URL of a translated page the UrlDecoder looks
for matches in the pages_overlays table. When it finds one, it
returns that overlay's uid as a page id, which is wrong.
It should return the overlay's pid instead.


fsuter:
I stumbled upon this issue when using a language-switching menu. The scenario is that I surf in the default language, then, at some point, click on the URL in another language. Depending on the page id returned, we either get to the wrong page or get a page not found error.

dmitryd:
Thank you!

PR #31, opened 2016-02-26
[TASK] Add "replace" section to composer.json

mbrodala:
This ensures that dependencies on these alternative package names can be resolved
in any case.

This is also essential for the TYPO3 composer installers to determine the extension 
name for installation via Composer vcs repository.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,10 @@
         "typo3",
         "realurl"
     ],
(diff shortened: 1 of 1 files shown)


dmitryd:
Thank you :)

mbrodala:
Thanks.

Issue #50, opened 2016-03-05
Segment lookup failing when Backend User logged in


bmoex:
I have an user secured environment in TYPO3 7.6, when running in 1.x version everything is working as intended. However 2.x seem to have some strange behaviour.

As I am on the website, it works as intended. However when I login the **backend** the frontend is giving segmentation errors. 
The error generating is: "Segment "dashboard" was not a keyword for a postVarSet as expected on page with id=1.". The page id for /dashboard/ should be 7. So there is a bit lookup failure.

The page setup is like this:
<img width="276" alt="screen shot 2016-03-05 at 12 05 19" src="https://cloud.githubusercontent.com/assets/2414665/13547238/d3f7572a-e2ca-11e5-9e8f-976e515494b1.png">

All menu separators have Exclude from speaking URL: Enabled.

bmoex:
Will try to debug more details

bmoex:
The problem lies in $pagesEnableFields and $disallowedDoktypes.
I locally changed the DOKTYPE_SPACER to DOKTYPE_SHORTCUT so this is 'bypassed'.

The pagesEnableFields adds the "(pages.endtime=0 OR pages.endtime>1457177640) AND (pages.fe_group='' OR pages.fe_group IS NULL OR pages.fe_group='0' OR FIND_IN_SET('0',pages.fe_group))" lookup.

The user object is not yet initiated in this stage and cause the user is logged in, the page segment is not read as intended. When you are not logged in the **backend**, it just reads the cache (which does work).

I will create a PR to 'fix this'. You can review it then!

dmitryd:
Spacers are not supposed to have subpages. Sorry, this will not be changed.

bmoex:
Hi Dmitry, where is this documented?

I figured this would lead to some discussion that's why i spliced both bugfixes in different PR's. You can decline the doktype PR then :-)

dmitryd:
1. https://www.google.com/search?q=typo3+spacer+page+type You will find multiple questions and answers concluding that it is not really possible to have subpages for spacers.
2. https://docs.typo3.org/typo3cms/TyposcriptReference/MenuObjects/CommonItemStates/Index.html, see SPC menu type and its description: "Spacers are pages of the doktype “Spacer”. These are not viewable pages but “placeholders” which can be used to divide menu items.". Just placeholders, not viewable. They are not really pages!

They are only visual separators. If you need subpages, use some other page type: folder, page-not-in-menu, etc.

[...]

PR #51, opened 2016-03-05
[BUGFIX] Ignore fe_group on enableFields()

bmoex:
User is not yet initiated at this stage

Resolves: #50


dmitryd:
Thank you! Good change!

PR #52, opened 2016-03-05
[BUGFIX] Allow DOKTYPE_SPACER to have subpages

bmoex:
Menu seperator can also be used as container for other pages

Resolves: #50


dmitryd:
Pages of type `Spacer` cannot have subpages. They are only visual separators.

Issue #72, opened 2016-03-14
config.tx_realurl_enable not evaluated for URLdecoding


liayn:
We used to have realurl installed on our dev machines as well, but disabled it by using TS like this:
```
[applicationContext = Development*]
  config.tx_realurl_enable = 0
[end]
```

realurl 2.0.9 still tries to decode URLs and fails (`#1420480928: RealURL was not able to find the root page id for the domain "kleindev"`) due to missing domain configuration.

In my opinion realurl must NOT take any action (not even looking at the configuration) if it is disabled.

dmitryd:
Not possible. When decoding runs, there is no TypoScript yet: TypoScript is processed for a specific page and page is not known yet (not decoded). That option works only when URL is encoded.

You have to either uninstall realurl or configure it correctly.

liayn:
Ok understandable, but not really funny of course.

It would be much more convenient to have a setting. What about an EM setting?

liayn:
Nevertheless I just wonder why realurl tries to fetch some config at all, since it should be able to detect that this is not a speaking url: `http://kleindev/76/index.php?id=165`

dmitryd:
No, I do not think so. If you installed realurl, than you did it on purpose. If you do not want to use it, just uninstall it. But better fix issues because you will face them in production anyway. How do you plan to go live with the project if you do not run it with full functionality daily? You actually should always have realurl enabled to spot any problems with urls.

dmitryd:
> Nevertheless I just wonder why realurl tries to fetch some config at all, since it should be able to detect that this is not a speaking url: http://kleindev/76/index.php?id=165

Actually this *is* a speaking URL :) But a completely wrong one.

liayn:
Multi-staging. On local dev, no realurl, on staging full realurl, on prod full realurl.
We cannot uninstall realurl for local dev, since it is required in our sitesetups, which ensures that all exts we use on production are installed automatically.

What is speaking here?? T3 simply runs in a subfolder of docroot.

dmitryd:
/76/ looks like a speaking URL to realurl.

liayn:
well, then the check is wrong. sitepath needs to be considered first!

liayn:
We talk about this line, so to say: https://github.com/dmitryd/typo3-realurl/blob/development/Classes/Decoder/UrlDecoder.php#L1078

dmitryd:
[...]

PR #74, opened 2016-03-14
[BUGFIX] Detect non-speaking URLs on decode

liayn:
Do not initialize realurl if it is not necessary at all. (#72)


dmitryd:
Thank you!

PR #78, opened 2016-03-15
[BUGFIX] Fix DataHandler hook cache handling

cedricziel:
The DataHandler - Hook issued a superfluous DELETE query that used
a non-existent field `uid`.

The call was removed completely since the cache backend handles the same
table entries in the line above.


dmitryd:
Thank you!

cedricziel:
That was fast - thank you :)

dmitryd:
PR was clear and well described – no reason to delay :+1:

PR #79, opened 2016-03-16
Only do the Decoding if no beUser present

Weedshaker:
The preview page would constantly overwrite L var's, which will not allow language switching while previewing. I guess it is safe to remove the decode when logged in. Else, it works as expected.

Cheers

### DIFF
--- a/Classes/Configuration/ConfigurationReader.php
+++ b/Classes/Configuration/ConfigurationReader.php
@@ -190,14 +190,16 @@ class ConfigurationReader {
 		$globalConfig = &$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl'];
 		foreach ($globalConfig['_DOMAINS']['decode'] as $domainName => $configuration) {
 			$checkThisConfiguration = false;
-			if ($domainName{0} == '/') {
-				// Regular expression, match only main host name
-				if (@preg_match($domainName, $this->hostName)) {
+			if($GLOBALS['TSFE']->beUserLogin < 1 || !$GLOBALS['TSFE']->beUserLogin){
+				if ($domainName{0} == '/') {
+					// Regular expression, match only main host name
+					if (@preg_match($domainName, $this->hostName)) {
+						$checkThisConfiguration = true;
+					}
+				}
+				elseif ($domainName === $this->hostName || $domainName === $this->alternativeHostName) {
 					$checkThisConfiguration = true;
 				}
-			}
-			elseif ($domainName === $this->hostName || $domainName === $this->alternativeHostName) {
-				$checkThisConfiguration = true;
 			}
 			if ($checkThisConfiguration) {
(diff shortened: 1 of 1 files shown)


Classes/Configuration/ConfigurationReader.php
```diff
@@ -190,15 +190,17 @@ protected function getConfigurationKeyByDomainDecode($configurationKey) {
 		$globalConfig = &$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl'];
 		foreach ($globalConfig['_DOMAINS']['decode'] as $domainName => $configuration) {
 			$checkThisConfiguration = false;
-			if ($domainName{0} == '/') {
-				// Regular expression, match only main host name
-				if (@preg_match($domainName, $this->hostName)) {
+			if($GLOBALS['TSFE']->beUserLogin < 1 || !$GLOBALS['TSFE']->beUserLogin){
```
dmitryd: This condition does not                                               make sense: both mean that beUserLogin are false.

dmitryd:
I believe this will break decoding if you have another tab open with BE on domain in non-default language.

Please, describe when this problem happens and what is actually wrong. "The preview page would constantly overwrite L var's" is not clear to me. I run tests with _DOMAINS active and have no issues.

Weedshaker:
The preview doesn't work to .jp (L=1) when logged in to the .com domain backend. Then while previewing, real url still overwrites the L vars, means L=1 can't be seen. Here my config:

``` php
	'_DOMAINS' => array(
        'decode' => array(
            'bla.com' => array(
                'GETvars' => array(
                    'L' => '0',
                ),
                'useConfiguration' => '_DEFAULT'
            ),
            'bla.jp' => array(
                'GETvars' => array(
                    'L' => '1',
                ),
                'useConfiguration' => '_DEFAULT'
            )
        ),
        'encode' => array(
            array(
                'GETvar' => 'L',
                'value' => '1',
                'useConfiguration' => '_DEFAULT',
                'urlPrepend' => 'http://bla.jp'
            ),
            array(
                'GETvar' => 'L',
                'value' => '0',
                'useConfiguration' => '_DEFAULT',
                'urlPrepend' => 'http://bla.com'
            )
        )
    )
```

This pull requests change solved my issue. We don't need to switch the domain for languages while someone is logged in. We could still open an other private window or a different browser. But maybe I am doing something wrong here. Shall I post the whole conf file?

dmitryd:
That pull request will create inconsistent behavior: realurl will work one way when BE user is logged in and anoter way when the user is not logged in. People will complain :smile:

I will fix it in a different way. The issue will be solved in release 2.0.10, which is due today.

Weedshaker:
Totally agree. This pull request has just been a quick fix... possible without looking deep into the extension. Anyways, I am going to try the new release out and give you some feedback on this issue.

PR #84, opened 2016-03-16
Performance tuning and Bugfixing

BoBaH83:
* make cacheImplementation before includeRealurlConfiguration so that developers can use their own by define the class in configuration file
* there is no need to lookup for alias_field directly in the table. in 99 percent it will fail, but is a performance killer if your tables are huge (mio. of rows)
* also there is no need to update the tstamp in tx_realurl_urlcache table on every page request multiple times

### DIFF
--- a/Classes/Cache/DatabaseCache.php
+++ b/Classes/Cache/DatabaseCache.php
@@ -139,15 +139,6 @@ class DatabaseCache implements CacheInterface, SingletonInterface {
 			$requestVariables = json_decode($row['request_variables'], TRUE);
 			// TODO Log a problem here because it must be an array always
 			$cacheEntry->setRequestVariables(is_array($requestVariables) ? $requestVariables : array());
-
-			// Update timestamp
-			$currentTime = time();
-			if ((int)$row['tstamp'] !== $currentTime) {
-				$this->databaseConnection->exec_UPDATEquery('tx_realurl_urlcache',
-					'uid=' . $this->databaseConnection->fullQuoteStr($cacheEntry->getCacheId(), 'tx_realurl_urlcache'),
-					array('tstamp' => $currentTime)
-				);
-			}
 		}
 
 		return $cacheEntry;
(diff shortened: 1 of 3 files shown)


Classes/Decoder/UrlDecoder.php
```diff
@@ -1225,7 +1204,7 @@ protected function searchForPathOverrideInPagesLanguageverlay($path) {
 				$result = GeneralUtility::makeInstance('DmitryDulepov\\Realurl\\Cache\\PathCacheEntry');
 				/** @var \DmitryDulepov\Realurl\Cache\PathCacheEntry $result */
 				$result->setLanguageId($this->detectedLanguageId);
-				$result->setPageId((int)$row['pid']);
+				$result->setPageId((int)$row['uid']);
```
BoBaH83: At this line the result pid is always 0 (zero) because of the select statement above.

Classes/Cache/DatabaseCache.php
```diff
@@ -177,15 +168,6 @@ public function getUrlFromCacheBySpeakingUrl($rootPageId, $speakingUrl) {
 			$requestVariables = json_decode($row['request_variables'], TRUE);
 			// TODO Log a problem here because it must be an array always
 			$cacheEntry->setRequestVariables(is_array($requestVariables) ? $requestVariables : array());
-
```
BoBaH83: I didnt find any relevant code, where you need the tstamp value. but you always make an update. why?

dmitryd:
> make cacheImplementation before includeRealurlConfiguration so that developers can use their own by define the class in configuration file

Cache implementation is not a part of realurl configuration. It should be defined in `ext_localconf.php` of the extension that provides the cache. Realurl configuration file should contain only realurl configuration.

Classes/Decoder/UrlDecoder.php
dmitryd: Integrated manually to the current development. Unfortunately I cannot take the rest because it removes functionality that either was there traditionally or made for future use.

dmitryd:
For future, make different pull requests for different parts of the code, please. I value contributions very much but it is really hard to merge when it is all at once.

Thanks for your help!

BoBaH83:
> Cache implementation is not a part of realurl configuration.

agree. I just have to make sure, my extension is loaded after realurl. It was just easy to maintain **all** things regarding realurl in the realurl-configuration-file

> For future, make different pull requests for different parts of the code

OK

> I cannot take the rest because it removes functionality that either was there traditionally or made for future use

OK. If I understand you right: there **IS** code, that is : 
- low-performance code just because it was implemented in v1
- useless now, but will be in future.

Well ... that means XClass for now and wait till it's useful in future :)


Just to show you: have a look at my database process list below

bildschirmfoto 2016-03-17 um 12 14 11

only real_url queries are visible :(
if there are > 500 users online at the same time, this queries are present in slow queries log. 
select, update, select, update, select update ...  and there is always an update, because $row['tstamp'] never equals time()

Thank You!

dmitryd:
Those queries shouldn't be slow. Check that your tables use InnoDB engine and not MyISAM. With MyISAM updates will be terribly slow but with InnoDB they will be ok.

PR #89, opened 2016-03-18
Make detectedLanguageId available in 404 case

christianfutterlieb:
On multilanguage websites, the "L" is configured in preVars very often. Because of this, when realurl comes to the most common 404 errors (not existing postVarSet), the preVars are decoded already.

Now, within my ext [pagenotfoundhandling](https://forge.typo3.org/projects/extension-pagenotfoundhandling), I want to be able to deliver the 404 page in the correct language. But when realurl throws the 404, I can't, because there is no "L" in $_GET.

The proposed solution would be a pretty straightforward one, I know. But the problem is gone for low cost and complexity ;-)

What do you think?

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -1376,6 +1376,8 @@ class UrlDecoder extends EncodeDecoderBase {
 	 */
 	protected function throw404($errorMessage) {
 		// TODO Write to our own error log here
+		// Register the detectedLanguageId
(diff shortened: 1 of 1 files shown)


dmitryd:
I see the point, it is useful. However, registers.... May be,` $_GET['L'] = $this->detectedLanguage`?

christianfutterlieb:
> May be, `$_GET['L'] = $this->detectedLanguage`?
Yes, sure. That's an even better and more logical solution than the (urghh) registers ;-)

dmitryd:
Did it :)

christianfutterlieb:
Great, thanks!

PR #99, opened 2016-03-25
Fix pagetitle "0" preventing encoding/decoding.

whefter:
Pages with title "0" cannot be encoded/decoded (and thus, not viewed) due to a not-type-safe if () comparision.

Our T3 still uses realurl 1.x due to compatibility issues that we haven't gotten around to fixing. Perhaps someone who knows where the relevant code lines are in the 2.x line can quickly insert the fix there, if necessary.


dmitryd:
Thank you!

Please, note that no new versions of realurl 1.x will be released!

PR #108, opened 2016-03-31
Handle follow up shortcut pages.

weber-ebusiness:
Hi,

we faced a bug using the new realurl version 2.x..

We often use a page tree which is started with some follow up shortcut pages, the first is the root page and the next is the page which is excluded from speaking URL, bundles different site pages and finally leads to the homepage.

Now we got an endless loop in an uncached situations. After some debuggings we find out, that the target page will always be the second shortcut and not the homepage which is the target of the second shortcut. 

The patch fixes our situation. The source comes originally from TYPO3\CMS\Frontend\Page\PageRepository->checkValidShortcutOfPage . 

Maybe there is a better solution that handles that case or there are some edge cases we don’t catch?! Please check that case and verify that patch.

Thx.

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -499,7 +499,68 @@ class UrlDecoder extends EncodeDecoderBase {
 			$this->throw404('Cannot decode "' . implode('/', $pathSegments) . '"');
 		}
 
-		return $result ? $result->getPageId() : ((int)$currentPid === (int)$this->rootPageId ? $currentPid : 0);
+		$targetPid = $result ? $result->getPageId() : ((int)$currentPid === (int)$this->rootPageId ? $currentPid : 0);
+		$return = $this->handleShortcutTargetPage($targetPid);
+
+		return $return;
+	}
+
+	/**
+	 * Tries to find shortcut page for targetPid
+	 *
+	 * @param int $targetPid
+	 * @return int
+	 */
+	protected function handleShortcutTargetPage($targetPid) {
+
+		$targetPage = $this->pageRepository->getPage($targetPid);
+
+		$dokType = (int)$targetPage['doktype'];
+		$shortcutMode = (int)$targetPage['shortcut_mode'];
+
+		if ($dokType === PageRepository::DOKTYPE_SHORTCUT && ($targetPage['shortcut'] || $shortcutMode)) {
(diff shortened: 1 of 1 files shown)


dmitryd:
Thank you, this is needed and useful! Most likely, this will come to 2.0.12. I won't be able to test it for 2.0.11 due to lack of time. But I will do it for 2.0.12!

dmitryd:
Could you post a graphical version of your page tree? I need to reproduce the issue first. Pseudographics will do. Something like:

|- Root (shortcut to ...)
|-- Subpage (excluded from url, ...)
|---- Subsubpage

dmitryd:
Might be related to #120

weber-ebusiness:
Hi,

yes, it seems that issue #120 is related. Maybe the patch fixes that too.

Our tree setup is like this:
|- Root (Exclude from speaking URL=1, Use as Root Page=1, Shortcut to "First subpage of selected/current page" => Pagetree page)
|-- Pagetree (Exclude from speaking URL=1, Shortcut to "First subpage of selected/current page" => Home page)
|--- Home

dmitryd:
This is extra tricky for RealURL. You have to understand that TYPO3 has no support for speaking URLs and creating speaking URLs is a very complicated business.

RealURL requires that your root page of the site (=Home) has "Use as Root Page" flag. RealURL starts searching from the root, so technically in your case it should start from "Home" because this is what your real root page is. While TYPO3 allows you to have arbitrary page tree configuration, RealURL cannot support any kind of structure. It is just not possible to make it work with every possible structure that human mind can create.

Make it simple and it will work.

It is best that you restructure your site to mark your "Home" page as a root page if you use RealURL.

dmitryd:
The patch in #120 will solve the issue for you but it will create problems with SEO for every other web site that uses RealURL and shortcuts (duplicate content). Therefore I am not happy about fixing this particular issue. This can be solved by simplifying page tree structure and making it more RealURl-friendly.

weber-ebusiness:
Hi,

I just test the newest dev branch and now surprisingly it works without any patches?!. I just can’t point out what really solves the problem, but it must be any latest changes after the patch.

So this issues could be closed. Thanks.

dmitryd:
Difficult to say, the log is very large between 2.0.11 and 2.0.12. But anyway, good that it works now.

PR #112, opened 2016-04-02
fix wrong/default language value on new cache entry

BoBaH83:
Description/How to reproduce:
* Multilanguage Site
* there is absolutely NO cache entry (tx_realurl_pathcache, tx_realurl_urlcache)
* user get linked from e.g. google on specific page in non default languge
* real url get the right path to page, but save the path in default language (zero)
Result is:
* the saved path is now used to make URL for default language
Example:
http://www.example.com/en/contact.html
switch to default language results to:
http://www.example.com/de/contact.html instead of http://www.example.com/de/kontakt.html


dmitryd:
Thank you!

PR #133, opened 2016-04-15
Remove field `tstamp` UrlEncoder.php

Ulfried:
In File Classes/Encoder/UrlEncoder.php the $insertArray contains the field `tstamp`, but ext_tables.sql doesn't define any `tstamp` field. So I think line 1139 (related to commit ed9086ae) should be deleted.


dmitryd:
Thank you!

PR #142, opened 2016-04-20
[FEATURE] added a hook for post-processing the configuration 

p2media:
We needed a simple hook to modify the configuration

### DIFF
--- a/Classes/Configuration/ConfigurationReader.php
+++ b/Classes/Configuration/ConfigurationReader.php
@@ -140,6 +140,15 @@ class ConfigurationReader {
 	}
 
 	/**
+	 * Returns the current mode.
+	 *
+	 * @return int
+	 */
+	public function getMode() {
+		return $this->mode;
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


BenjaminBeck:
Why not use Signal/Slot mechanism?
- http://stackoverflow.com/questions/30544422/typo3-extbase-how-to-use-core-signal-slots
- https://forge.typo3.org/issues/59089

dmitryd:
@BenjaminBeck RealURL uses traditional hooks in other part due to compatibility. Mixing traditional hooks with signals would not be a good idea.

BenjaminBeck:
Ok, Thank your for the explanation

dmitryd:
I think you would also want to know what mode are you running in: decode or encode. Probably adding a getter for the `ConfigurationReader::$mode` make sense.

p2media:
Yes, that was a dumb idea 😅  
I've removed `mode` from the parameters and added a proper getter function `getMode()`

dmitryd:
Thank you for your work!

PR #157, opened 2016-05-12
Fixed bug in auto configurator, selecting wrong field for lang iso code

hoermannklaus:
When creating the auto configuration for realurl there is a bug in the language select query, where the uid is falsly selected for the language_isocode.

### DIFF
--- a/Classes/Configuration/AutomaticConfigurator.php
+++ b/Classes/Configuration/AutomaticConfigurator.php
@@ -83,7 +83,7 @@ class AutomaticConfigurator {
 		if ($this->hasStaticInfoTables) {
 			$languages = $this->databaseConnection->exec_SELECTgetRows('t1.uid AS uid,t2.lg_iso_2 AS lg_iso_2', 'sys_language t1, static_languages t2', 't2.uid=t1.static_lang_isocode AND t1.hidden=0');
 		} else {
-			$languages = $this->databaseConnection->exec_SELECTgetRows('t1.uid AS uid,t1.uid AS lg_iso_2', 'sys_language t1', 't1.hidden=0');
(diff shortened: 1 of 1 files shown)


Classes/Configuration/AutomaticConfigurator.php
```diff
@@ -83,7 +83,7 @@ protected function addLanguages(array &$configuration) {
 		if ($this->hasStaticInfoTables) {
 			$languages = $this->databaseConnection->exec_SELECTgetRows('t1.uid AS uid,t2.lg_iso_2 AS lg_iso_2', 'sys_language t1, static_languages t2', 't2.uid=t1.static_lang_isocode AND t1.hidden=0');
 		} else {
-			$languages = $this->databaseConnection->exec_SELECTgetRows('t1.uid AS uid,t1.uid AS lg_iso_2', 'sys_language t1', 't1.hidden=0');
+			$languages = $this->databaseConnection->exec_SELECTgetRows('t1.uid AS uid,t1.language_isocode AS lg_iso_2', 'sys_language t1', 't1.hidden=0');
```
liayn: `language_isocode` is not present in TYPO3 6.2, which this ext claims compatibility for.

Core-wise we have available in
6.2: static_lang_isocode, which points to a static_languages records if present
7.6: language_isocode holds the iso-code directly (always), static_lang_isocode is there for backwardscompat when SIT is installed

liayn: The whole logic should be changed to:

If CMS 7, use `language_isocode` unless empty, otherwise fall back to 6.2 handling.
If CMS 6.2, if SIT is available and `static_lang_isocode` is set, resolve the relation.

liayn:
@dmitryd Did you take my comment from above into account?

Issue #168, opened 2016-05-20
No possiblity to retrieve getValue in encodeTitle_userProc


liayn:
Formerly the member `orig_paramKeyValues` was publicly available and could be used in the `encodeTitle_userProc` hook to retrieve the original value of a GET parameter.
All members of the UrlEncoder are now protected, which is good, but now there is no way to retrieve the GET parameter, where the `$processedTitle` originated from.

I propose to hand over this information, as well as the language information, to the `cleanUpAlias` method as well as the hook.

I'll push a PR.

dmitryd:
See #150.

liayn:
Yes and no... the language information is still not there.

liayn:
But I don't insist on this for the time being. I can live with #150 of course. Thanks.

liayn:
When will you release a new version which includes #150?

dmitryd:
Another accessor can be added. Just write what you need.

dmitryd:
No plans yet but you can use dev version for now.

liayn:
Language is post-processed in `\DmitryDulepov\Realurl\Encoder\UrlEncoder::createAliasForValue` so, an accessor is not sufficient in that case.

But in terms of accessor, it would be nice to have `\DmitryDulepov\Realurl\Utility::convertToSafeString` available in this hook. Unfortunately it is not static, even though it has no dependencies on members. Moreover the `strtolower` should be controllable via parameter.
May I push a PR for that?

dmitryd:
There is a `$this->getUtility()` method now. Or you can instantiate `Utility` class.

> Moreover the `strtolower` should be controllable via parameter.

You can but for the external use of the function only. Default behavior should stay as it is. RealURL should lowercase URLs as it always did.

liayn:
sure!

PR #169, opened 2016-05-20
[TASK] Make strtolower optional for convertToSafeString

liayn:
Fix: #168

### DIFF
--- a/Classes/Utility.php
+++ b/Classes/Utility.php
@@ -63,12 +63,15 @@ class Utility {
 	 * Converts a given string to a string that can be used as a URL segment.
 	 * The result is not url-encoded.
 	 *
-	 * @param string $string
+	 * @param string $processedTitle
 	 * @param string $spaceCharacter
+	 * @param bool $strToLower
 	 * @return string
 	 */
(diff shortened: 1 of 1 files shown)


Classes/Utility.php
```diff
@@ -65,9 +65,10 @@ public function __construct(ConfigurationReader $configuration) {
 	 *
 	 * @param string $string
 	 * @param string $spaceCharacter
+	 * @param bool $strToLower
 	 * @return string
 	 */
-	public function convertToSafeString($string, $spaceCharacter = '-') {
+	public function convertToSafeString($string, $spaceCharacter = '-', $strToLower = true) {
 		$processedTitle = $this->csConvertor->conv_case('utf-8', $string, 'toLower');
```
dmitryd: Missing `if` here too?

liayn: Good point. Why is the strtolower below necessary?

dmitryd: Because of `specCharsToASCII` on line 74.

This code is historical, from the old version. I am not sure why there is this double conversion but I do not want to change it without a real reason.

dmitryd:
Thank you :)

PR #179, opened 2016-06-08
[BUGFIX] Fix #175: Allow same pathsegment for different languages

### LINKED ISSUE #175: Allow same pathsegment for different languages
Currently the `tx_realurl_pathsegment` field of the `pages_language_overlay` table has the `eval`-configuration uniqueInPid. This means, it is not possible to use the same pathsegment for different languages. But it is possible to have the same segment in different (similar) languages. Then they will get a number appended. Example:
en/login/
de/login0/

I think it is ok, to drop the `eval`-field and allow the same segment for multiple page-translations (with the assumption, that a page only has translations with different `sys_language_uid`).

rvock:
Allow the same `tx_realurl_pathsegment` within the same pid for `pages_language_overlay`.  Different entries will still be different because of their `sys_language_uid`


dmitryd:
Thank you!

PR #191, opened 2016-07-04
Update DatabaseCache.php

### LINKED ISSUE #162: Shorcut page subpages leads to 404 page
In typo3 v7.6.6 and realurl v2.0.14 I have same problem like here #42 
But only on localized pages
When I loged off from BE it works

WorksDev:
Fixes #162, we need a fallback if no Language Var is set otherwise we get more and more cacheentrys with the same fields wich causes the 404.


dmitryd:
Please, test the recent development branch. I believe this is already fixed in a different way.

WorksDev:
Hi, 
think it is working fine right now. Thanks for your work.

Issue #190, opened 2016-07-04
Fluid form validation leads to HMAC validation error when using realurl on TYPO3 6.2.x


jacobsenj:
I've got a fluid form, which does add some 
<details><summary>hidden validation parameters.</summary>
```html
<div>
...
<input type="hidden" name="tx_extension_pi2[__trustedProperties]" value="a:4:{s:6:&quot;filter&quot;;a:4:{s:9:&quot;continent&quot;;i:1;s:7:&quot;country&quot;;i:1;s:7:&quot;vehicle&quot;;i:1;s:5:&quot;brand&quot;;i:1;}s:6:&quot;action&quot;;i:1;s:6:&quot;sortBy&quot;;i:1;s:9:&quot;sortOrder&quot;;i:1;}2eb7360e14bed59fc5db85eddceb1bfc9d5e7651">
</div>
```
</details>
A part of the hidden validation parameters is an array holding keys like __trustedProperties which is a serialized array of the allowed form fields.

When disabling realurl my form will show this string value in typo3/sysext/extbase/Classes/Security/Cryptography/HashService.php:64:
```
a:4:{s:6:"filter";a:4:{s:9:"continent";i:1;s:7:"country";i:1;s:7:"vehicle";i:1;s:5:"brand";i:1;}s:6:"action";i:1;s:6:"sortBy";i:1;s:9:"sortOrder";i:1;}
```

if I enable realurl the following contents are shown there:
```
a:4:{s:6:\"filter\";a:4:{s:9:\"continent\";i:1;s:7:\"country\";i:1;s:7:\"vehicle\";i:1;s:5:\"brand\";i:1;}s:6:\"action\";i:1;s:6:\"sortBy\";i:1;s:9:\"sortOrder\";i:1;}
```

which then leads to the error message 
`#1320830018: The given string was not appended with a valid HMAC`

I tracked everything down to the use of `$_GET` in ext:realurl. It seems to be the source of the problem because I also found this code in
<details>
<summary>typo3/sysext/core/Classes/Utility/GeneralUtility.php _GET() method</summary>
```php
		// Removes slashes since TYPO3 has added them regardless of magic_quotes setting.
		if (isset($value)) {
			if (is_array($value)) {
				self::stripSlashesOnArray($value);
			} else {
				$value = stripslashes($value);
			}
		}
		return $value;
```
</details>

and then modified the usage of `$_GET` in
<details>
[...]

PR #198, opened 2016-07-17
Use GeneralUtility::_GET() instead of $_GET

helhum:
TYPO3 until TYPO3 version 6.2 adds slashes to values of super global $_GET
To avoid slashes added twice, the GeneralUtility::_GET() API method
must be used to obtain GET variable values.


dmitryd:
Related to #190.

dmitryd:
Thank you!

helhum:
Thanks for the quick merge

PR #208, opened 2016-07-25
Fix language support for the same uri

edit-michael:


dmitryd:
This will fail with _DOMAINS.

dmitryd:
Rejected.

PR #211, opened 2016-07-27
[FEATURE] tx_realurl_pathoverride also used for urls of subpages

franzkugelmann:
if tx_realurl_pathoverride is set, the url the the very page where it is set is reduced to the value given in tx_realurl_pathsegment. However urls to subpages are still generated with the full url, ignoring tx_realurl_pathoverride. This commit changes that. Also subpages start with the short path given in tx_realurl_pathsegment. 
The UrlDecoder already can handle these urls correctly.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -483,6 +483,14 @@ class UrlEncoder extends EncodeDecoderBase {
 					unset($overlay);
 				}
 			}
+			// if pathoverride is set, use pathsegment also for all subpages to shorten the url and throw away all segments found so far
+			if ($page['tx_realurl_pathoverride'] && !empty($page['tx_realurl_pathsegment'])) {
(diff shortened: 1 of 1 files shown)


dmitryd:
No, because this will change the existing behavior and potentially render many existing URLs on many sites in a different way breaking site ranks. Any change you make may not impact compatibility in such a big way.

If you want this feature, it has to be enabled by a separate option in the configuration.

franzkugelmann:
Agreed, downwards compatibility is very important here.
If you find this feature useful as we do, would you like to implement the configuration option as you see fit? We would sponsor it. No hurry needed, enjoy your holidays first :)

dmitryd:
Yes, the feature is useful and I would love to see it in RealURL. I will be back after August 8th, and we can discuss :)

dmitryd:
There were some extra work and more changes:

1. The original branch got old and there were many merge conflicts.
1. If I use override like `override/test` it was correct only for the page where it was set but all other pages got `overridetest` (without slash), so I fixed this too.
1. Finally, some other merge conflicts due to parallel development.

No option to turn this on. I think we can treat it as a fixed bug.

Issue #152, opened 2016-05-03
Realurl section in 'Edit Page' disappeared after Typo3 update (TYPO3 7.6.6 only!)


thomasmstein:
Dmitry,
I noticed that after an update from Typo3 7.6.5 to 7.6.6 the "**Speaking URL path segment:**" and "**Exclude from speaking URL:**" sections disappeared from the bottom of the 'Edit Page' (while the RealURL administration module remains accessible)
I am running Typo3 with bootstrap_package 7.0.2.
Is there a way to reactivate that section.
Thanks
Thomas

![edit_page_no_realurl](https://cloud.githubusercontent.com/assets/17682916/14972857/17d6df56-1124-11e6-9b3f-2cc0659f8e9a.png)

schnappenberger:
You wil find it in Tab Categories ...

dmitryd:
Patch version is supposed not to break things or introduce incompatible changes. Looks like it is not the case with TYPO3 7.6.6. There are too many reports about broken things there.

thomasmstein:
Frank and Dmitry,
Many thanks for the hint and quick response. I presume that the Speaking URL section was not supposed to be ending up under the Category tab ... ;-)  ?
Cheers
Thomas

wouter90:
This is a core bug which is fixed already and will be pary of 7.6.7

PR #213, opened 2016-07-28
Make "Disable RealURL cache" visible on page record edit form

cweiske:
In the TYPO3 edit page record page, the field `tx_realurl_nocache`
was not visible on TYPO3 7.6.


dmitryd:
What TYPO3 version exactly is it? To me it looks like #152.

cweiske:
TYPO3 7.6.8

dmitryd:
Not needed currently because this flag is not in use anyway.

Issue #218, opened 2016-07-30
Non-Breaking-Spaces are removed from url


rvock:
If a page title contains a non-breaking-space (ALT+Space on OS X), it will not be replaced with a - like a normal space but removed completely.

Example:
`page 1` will be encoded as `page-1`
`page 2` (note the nbsp between page and 2) will be encoded as `page2`

The Utility::convertToSafeString can easily be adjusted to respect all space-characters (nbsp, tab, normal space):
```php
// the space in the next line is a non-breaking space
$processedTitle = preg_replace('/[ \s\-+_]+/u', $spaceCharacter, $processedTitle);
```

dmitryd:
Note: breaks compatibility.

rvock:
Hm, yes, it breaks compatibility. If a page title had a non-breaking-space before, the url would not contain it. If this url is saved or linked somewhere, this old link would not work after this change.

I think, this would only affect very few pages, because a non-breaking-space will nearly never appear in a title. On windows, you only can input it with copy&paste or with ALT+0-1-6-0 . On OS X you can input it with Alt+Space.

If you want backwards compatibility, I could adjust the createPathCacheEntry-method in UrlDecoder:
https://github.com/dmitryd/typo3-realurl/blob/25f1ba4c454aa017b1c0c461792a8512c5b5c202/Classes/Decoder/UrlDecoder.php#L343

This method could also check if the url would match `segment` with nbsp removed...

rvock:
You're right, it was a normal space. I had the hex-codes mixed up.

dmitryd:
See #219

Issue #207, opened 2016-07-25
Handling of page title "0"


co-operate:
RealURL seems not to be able to handle URLs to pages with a page title "0" (zero).  

The URL `domain.com/some/path/0/` is generated, but the link will actually only lead to `domain.com/some/path/`.

PR #220, opened 2016-07-30
[FEATURE] Base for Testcases (based on configuration of extension_builder)

rvock:
I created some basic functional testcases for realurl. I've based the configuration for the tests on code from [extension_builder](https://github.com/TYPO3-extensions/extension_builder).

To run the test-cases, you need to install the dependencies using composer:
```
composer install --dev
```

Afterwards, you can run the tests using `composer test`. You need a database user `realurl` without a password to run the tests. Alternatively you can run the following commands:
```
export TYPO3_PATH_WEB=`pwd`/.Build/Web
export typo3DatabaseName='realurl'
export typo3DatabaseHost='127.0.0.1'
export typo3DatabaseUsername='realurl'
export typo3DatabasePassword=''
.Build/bin/phpunit --colors -c .Build/vendor/typo3/cms/typo3/sysext/core/Build/UnitTests.xml Tests/Unit/
.Build/bin/phpunit --colors -c .Build/vendor/typo3/cms/typo3/sysext/core/Build/FunctionalTests.xml Tests/Functional/
```

The tests generate a complete TYPO3 installation within the `.Build/Web` folder and creates test instances for each test-class in `.Build/Web/typo3temp/var/tests/`

Currently there are only some simple tests for URL encoding:
- normal pages and child pages
- configuration of tx_realurl_exclude, tx_realurl_pathoverride and alias

I also included tests for some of the latest bug reports:
- #218 
- #207 

In the future I would like to add tests for encoding more complicated URLs:
- various language configurations
- Parameters in the URL
- multiple domains

I also added a .gitattributes file to exclude the testcases from the zip-releases here on github. I do not know how the releases for TER are generated, but they also could exclude the tests. But they are not large so it shouldn't matter much.

Ping #8

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+/.gitattributes export-ignore
+/.travis.yml export-ignore
+/Tests/ export-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.Build/*
+vendor/*
+composer.lock
--- a/Tests/Fixtures/realurl.xml
+++ b/Tests/Fixtures/realurl.xml
@@ -1,0 +1,119 @@
+<?xml version="1.0" encoding="utf-8"?>
+<dataset>
+	<sys_domain>
+		<uid>1</uid>
+		<pid>1</pid>
+		<domainName>www.example.com</domainName>
+	</sys_domain>
+<!--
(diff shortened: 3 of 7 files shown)


dmitryd:
Thanks!

PR #225, opened 2016-08-09
[BUGFIX] Overwrites autoconf file

spoonerWeb:
In a situation where the autoconf file could not be deleted due to a change in sys_domain table, the autoconf will be appended to the file.
This result will lead to a 500 page.

### DIFF
--- a/class.tx_realurl_autoconfgen.php
+++ b/class.tx_realurl_autoconfgen.php
@@ -83,7 +83,7 @@ class tx_realurl_autoconfgen {
 		$fileName = PATH_site . TX_REALURL_AUTOCONF_FILE;
 
 		$lockObject = $this->apiWrapper->getLockObject($fileName);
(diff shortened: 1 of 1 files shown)


class.tx_realurl_autoconfgen.php
```diff
@@ -83,7 +83,7 @@ public function generateConfiguration() {
 		$fileName = PATH_site . TX_REALURL_AUTOCONF_FILE;
 
 		$lockObject = $this->apiWrapper->getLockObject($fileName);
-		$fd = @fopen($fileName, 'a+');
+		$fd = @fopen($fileName, 'w+');
 		if ($fd) {
 			// Check size
 			fseek($fd, 0, SEEK_END);
```
dmitryd: With w+ this line is not needed. Please, check the code and clean up further.

dmitryd:
This is for v1 - will not be updated anymore.

PR #242, opened 2016-09-07
[BUGFIX] preVars valueDefault has wrong behavior

monosize:
In realurl version 1.x a defined "valueDefault" was added to the uri, when the urlParameter didn't exists.
Since version 2.x the valueDefault is ignored.

```php
'preVars' => array(
    array(
        'GETvar' => 'L',
        'valueMap' => array(
            'de' => 0,
            'fr' => 1,
        ),
        'valueDefault' => 'de',
    ),
),
```

Example: urlParameter L is not set.

in Version 1.x the 'valueDefault' was added to the url: ```http://www.domain.com/de/pagename```
since version 2 the 'valueDefault' is ignored: ```http://www.domain.com/pagename```

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -545,7 +545,7 @@ class UrlEncoder extends EncodeDecoderBase {
 	protected function encodePreVars() {
 		$preVars = (array)$this->configuration->get('preVars');
 		if (count($preVars) > 0) {
-			$segments = $this->encodeUrlParameterBlock($preVars);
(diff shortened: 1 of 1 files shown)


dmitryd:
RealURL 2.x tries to be compatible as much as possible with 2.x but still it is not 100% compatible. Since it is a different product, some people already adapted to the current behaviour and merging this PR will break it. So I will not merge it. Sorry :(

To solve your issue with language preVar, please, read [this](https://github.com/dmitryd/typo3-realurl/wiki/Notes-for-Integrators#configuring-languages).

monosize:
Ok. I  close this PR

Issue #245, opened 2016-09-09
Wrong content delivered when chashcache is empty and fixedPostVars-URLs are accessed


cweiske:
## Problem
Realurl 1.13.6 stores invalid entries in the url decode cache when the chashcache has no entry for the current URL.

This leads to wrong content delivered for those URLs (often the same content, since all get the same `newHash`).

## Situation
Plugin that pages data, and a realurl configuration that maps the necessary GET parameter into the URL:
```
$TYPO3_CONF_VARS['EXTCONF']['realurl']['_DEFAULT'] = [
    'init' => [
        'enableCHashCache'     => true,
        'enableUrlDecodeCache' => true,
        ...
    ],
    ....
    'fixedPostVars' => [
        23 => [
            ['GETvar' => 'page'],
        ]
    ],
    ...
];
```
This gives us paged URLs for the `news` page 23:

URL | Decoded URL
--- | -----------
`/news/` | `/?id=23&page=`
`/news/2/` | `/?id=23&page=2`
`/news/3/` | `/?id=23&page=3`

## Triggering the error

1. Empty frontend cache
2. Empty all realurl cache tables (`TRUNCATE TABLE tx_realurl_chashcache; TRUNCATE TABLE tx_realurl_urldecodecache; TRUNCATE TABLE tx_realurl_urlencodecache;`)
3. Open URL `/news/`. The correct content is delivered since the FE cache is empty. Since page 2 is linked on the page, a `chashcache` record is created for it.
4. Open URL `/news/2/`. Since there is a `chashcache` entry for it, the correct content is delivered
5. Open URL `/news/4/`. There is no `chashcache` entry for it because TYPO3 did not yet generate a page that links to it. The content of `/news/` is shown!

This is because realurl does not add a `cHash` parameter to the GET variables,  and `TypoScriptFrontendController` internally ignores the GET variable and thus generates the same `newHash` as it does for `/news/`. Thus it delivers the content of `/news/`.

The problem gets worse because realurl creates an entry for `/news/4/` in the `urldecodecache`, without the `cHash` parameter. This means that - even if in the meantime another page call leads to a `chashcache` entry for `/news/4/` - it will never be used because the url decode cache does not contain it.

## Migitation
- At first, realurl should not store entries in `urldecodecache` if they have GET parameters with values and no `cHash`.
-- This prevents the problem that the wrong page content is shown even if in the meantime a correct `chashcache` entry for the URL is generated
[...]

PR #246, opened 2016-09-09
Better cope with missing chashcache entries

cweiske:
The problem is fully described in
https://github.com/dmitryd/typo3-realurl/issues/245

Resolves: #245

### DIFF
--- a/class.tx_realurl.php
+++ b/class.tx_realurl.php
@@ -1762,6 +1762,16 @@ class tx_realurl extends tx_realurl_baseclass {
 
 
 				if (!$this->isBEUserLoggedIn() && $this->canCachePageURL($cachedInfo['id'])) {
+					if (isset($cachedInfo['GET_VARS']) && count($cachedInfo['GET_VARS'])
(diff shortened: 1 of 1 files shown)


dmitryd:
No, this is not a correct solution. Realurl only encodes and decodes URLs. It should not handle cache issues.

cweiske:
> Realurl only encodes and decodes URLs. It should not handle cache issues.

But who if not realurl is responsible for the cache tables that realurl fills?

dmitryd:
> But who if not realurl is responsible for the cache tables that realurl fills?

That's right. And you should not go and truncate them because it is realurl's job, not yours :) Would you like if somebody comes to your computer and wipes your documents without a warning (like bank access codes) and than demand to recover from it? Probably not. Realurl tables belong to realurl. It is valuable data and it should not be manipulated by anybody in any way.

cweiske:
I think we can stop discussing now since we have different views on this issue. Thanks for maintaining realurl.

PR #261, opened 2016-09-19
check for pages with override if rootpage_id is inside rootline

### LINKED ISSUE #259: Page with Override Whole path segment leads to infinite redirect
I have the following page tree for 1 domain:

|- Customer ABC (uid 1, normal page)
|-- Domain ABC (uid 2, set as root page, contains correct domains)
|--- "Shortcut to first subpage" (uid 3, exclude from speaking URL)
|---- First Subpage (uid 4, override whole page path with "Shortcut to first subpage")

In the realconf configuration rootpage_id is set to 2 (integer).

A link to the pages with uid 3 and 4 would be the same  -> /shortcut-to-first-subpage
So realurl should determine that the link /shortcut-to-first-subpage is always decoded to page 4 because of the override feature. But instead it determins page 3 as correct. This leads to a infinite redirect as TYPO3 determins that page 3 is a shortcut and should be redirected to page 4 but realurl determins for that URL that the page must be 3 instead 4 which than again triggers the redirect of TYPO3 and so on.

This is because the configured rootpage_id is 2 but UrlDecoder::searchForPathOverrideInPages checks only if the first page in the rootline of page 4 is the same as the configured root page, which it's not in this case. In realurl 1.x it was checked if the rootpage_id was inside the rootline and not if the first page in the rootline matches the configured rootpage_id.

So I'm wondering if this is a bug or intended behaviour. Because this would mean that a root page of a domain always has to be on the root level in the page tree and not like in my scenario.

P.S. As soon as I set the configured rootpage_id to 1 the redirect is gone and the correct page is displayed but of course the URL is not the same as it contains now the segment for page 2. I could get around this by setting the page 2 to be excluded from the url but I'm still wondering if that's correct.

hannesbochmann:
before it was checked if the first page in the rootline of the page with override equals the configured rootpage_id.

fixes #259


dmitryd:
stop-page-tree

Doesn't this help?

hannesbochmann:
This would help but nevertheless breaks backward compatibility as action by the user is necessary to make everything work as before.

dmitryd:
Well, I understand that it is easier to invent a complex page structure and ask realurl to deal with it than to make a simple structure or set a checkbox to compensate for the complex page structure. But I am reluctant to spend time on it right now because it is not a typical situation. It is low priority because can be solved without code changes.

hannesbochmann:
Okay, I don't really understand why you don't want to merge this pull request right now but I guess I have to accept this.

dmitryd:
Because the release of 2.1.0 is near and I only add code that I am sure is stable and does not cause any side effects.

hannesbochmann:
That's understandable.

By the way, thanks for the great work.

hannesbochmann:
I checked again and setting the stop page tree didn't work.

adrienjacob:
I confirm the suggested commit would solve a similar issue I have... Cf. https://github.com/dmitryd/typo3-realurl/issues/357

dmitryd:
Merged manually. See #259.

PR #264, opened 2016-09-20
[BUGFIX] Prevent not-fully translated URLs to be stored

christianfutterlieb:
Mitigates a possible cache-flooding when calling many
URLs with random parameters

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -206,9 +206,24 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 	 */
 	public function storeCacheRecord() {
 		// If it is still not there (could have been added by other process!), than store it
-		if ($this->createdCacheEntry && !$this->isExpiredPath && !$this->getFromUrlCache($this->speakingUri)) {
+		if ($this->shouldStoreCacheRecord()) {
 			$this->putToUrlCache($this->createdCacheEntry);
 		}
+	}
+
+	/**
+	 * Returns true if $this->createdCacheEntry should be stored
+	 *
+	 * @return boolean
+	 */
+	protected function shouldStoreCacheRecord()
+	{
+	    return (
+	            $this->createdCacheEntry
+	            && !$this->isExpiredPath
+	            && !$this->getFromUrlCache($this->speakingUri)
+	            && strpos($this->speakingUri, '?') === false
+	    );
 	}
 
 	/**


christianfutterlieb:
Hi Dmitry

While checking out the code I noticed, that any URL passed to `UrlDecoder` will be stored to cache table `tx_realurl_urldata`.

A simple one-liner confirms that, the table got spammed with 10'000 useless entries

    for i in {1..100} ; do for j in {1..100} ; do curl "http://myhost/page/?test%5Bparam1%5D=$i&test%5Bparam2%5D=$j" > /dev/null ; done ; done

Because there is no way to know which URLs are legitimate and can be stored, it could be a good idea to store fully translated URLs only. This would "verify" that

1. the URL is valid (otherwise `storeCacheRecord()` would not have been called)
2. the parameters are known to the configuration

Just for clarification: this applies to `UrlDecoder` only. Other URLs generated by `UrlEncoder` are not meant to be part of that logic. The difference is that the decoder reacts on user input and the encoder on (hopefully correct) configuration.

Regards, Christian

dmitryd:
Yes, this is correct. The encoder does not not have knowledge if the URL is legitimate but it saves it anyway because if such URL has to be created by the encoder, it will reuse existing URL from the cache.

Current behavior is correct. This pull request will not be implemented unless I miss something big here.

In general, open the issue first and ask. Spare me time checking the code and explaining all these details...

christianfutterlieb:
> Yes, this is correct. The encoder does not not have knowledge if the URL is legitimate but it saves it anyway because if such URL has to be created by the encoder, it will reuse existing URL from the cache.

As I explained it is not about the encoder but the decoder.

> Current behavior is correct.

Really? The part with the possible cache-flood too?

dmitryd:
I afraid you know about realurl slightly less than I do and do not understand relations between encoder & decoder and why the decoder saves URLs like that.

All decoded URLs must be saved to the cache. You can call it whatever you like but it is designed to be like that on purpose.

I do not plan to implement this change because it will cause more work for the decoder too, which means slowing down sites and complaints from users.

Please, do not try to fix something that you do not really understand. Encoder and Decoder share information and saying "it is not about the encoder but the decoder" means that you do not understand how realurl works.

This topic is closed for me. I cannot afford spending more time on this. Sorry. I have work to do.

christianfutterlieb:
I'm afraid too that you don't want to acknowledge the problem. Just forget about the patch, it's not about that piece of code, I'm talking about a security problem of realurl. I'll try to explain it in a way everybody can understand (and avoid talking about realurl internals that many people including me don't catch on to):

# Vulnerability 1: Cache Flooding in realurl cache

Realurl stores every URL it receives to a database table. Knowing that an attacker will be able to call an URL over and over again by changing arbitrary GET parameters in every request:

    https://example.org/page/?a=1
    https://example.org/page/?a=2
    https://example.org/page/?a=3
    ...

Realurl creates an entry in it's cache table in the database, even it's senseless. Like this the database table can be flooded with data leading in extremo to a system going down because of lack of storage space.

Filling the table would surely be more efficient when the produced URLs are much larger than in the example above.

# Vulnerability 2: Cache Flooding in TYPO3 pagecache

This risk applies to the development branch only. To exploit this issue, some conditions must be met:

1. TYPO3 configuration (this option is default):

    $GLOBALS['TYPO3_CONF_VARS']['FE']['pageNotFoundOnCHashError'] = true;

2. Realurl configuration (currently in development branch)

    'init/redirectOnChashError' = true;

3. The page must not have a cached version in TYPO3 pagecache (at the time of the attack)

When the attack method used in the description of the first vulnerability is run on a system that meets the above requirements, not only the realurl cache table gets flooded, but the TYPO3 pagecache too. This would be far more serious, because the page cache stores lots of data.

What happens: realurl re-creates the URL with typolink (which appends a correct cHash) and then redirects there. The page gets called with the correct cHash and will thus be cached in pagecache.

# Conclusion

I consider these to be serious security problems.

Now, @dmitryd, I hope you are willing to spend your valuable time on thinking about these vulnerabilities.

dmitryd:
No. If you think it is a vulnerability, write to security@typo3.org. But I will not be one, who "fixes" it because I will not be developing realurl for long due to people like you, who try to fix something without understanding what you are doing and how it is going to break realurl. I had enough. As soon as 2.1.0 is out, I will quit.

dmitryd:
> What happens: realurl re-creates the URL with typolink (which appends a correct cHash) and then redirects there. The page gets called with the correct cHash and will thus be cached in pagecache.

Have you actually tried that? :) I guess not. It creates such URL only if:
- there is no cHash
- the plugin requires cHash

Good luck to you and all others with realurl.

christianfutterlieb:
Glad, your about the topic, finally.

Yes, you're right, the pagecache flood only occurs with a plugin requiring a cHash, I add this to my previous post. But that is not the main problem as it is not very likely (although possible) to happen. The focus of the discussion should stay on the first problem.

I don't comment on your accusations or personal attacks. I know you've been doing really lot of oss stuff and I'm still a thankful user of TYPO3 and realurl. But in my opinion your statements are just a shame.

dmitryd:
Btw, this may get a different approach. Since the decoder cannot figure out if it decoded a URL fully correctly, it makes sense to use the url cache read-only in the decoder and not store anything. In case of a vlid url it will be re-created by the encoder anyway.

dmitryd:
https://github.com/dmitryd/typo3-realurl/compare/feature/removeUrlCacheFromDecoder

christianfutterlieb:
> .. use the url cache read-only in the decoder and not store anything ..

Good thorough solution, thanks.

Note: commit with the change: e2049d8baf9e9d004557248a0ce9e9cf0aafc9eb

PR #280, opened 2016-09-24
[TASK] In case of duplicated urldata entries, prefer those with cHash

webian:
It could happen that RealUrl encodes more than one urldata entry for the same speaking url. Usually one of the entries does have cHash and the other does not. This commit makes sure that, in this case, the entry with the cHash is taken into account over the one without.

### DIFF
--- a/Classes/Cache/DatabaseCache.php
+++ b/Classes/Cache/DatabaseCache.php
@@ -199,6 +199,15 @@ class DatabaseCache implements CacheInterface, SingletonInterface {
 				'', 'expire'
 		);
 
+		// More than one urldata entry? Then prefer those with cHash
+		if (count($rows) > 1) {
+			if (preg_match('/cHash/', json_encode($rows))) {
+				$rows = array_filter($rows, function ($row) {
+					return strpos($row['original_url'], 'cHash') !== false ? true : false;
+				});
+			}
+		}
+
 		$row = null;
 		foreach ($rows as $rowCandidate) {
 			if (is_null($languageId)) {


webian:
@dmitryd please check this PR.
In my production sites I often have duplicated urldata entries and the one without cHash comes before the one with (lower uid).
It happens that if is_null($languageId) in getUrlFromCacheBySpeakingUrl() then Realurl considers the first entry without cHash and then TYPO3 display the wrong page.
This commit should mitigate this bad behavior.

webian:
This SQL can be useful to check for duplicated urldata entries:
`SELECT * FROM tx_realurl_urldata a JOIN ( SELECT speaking_url FROM tx_realurl_urldata GROUP BY speaking_url HAVING count(*) > 1 ) b ON a.speaking_url = b.speaking_url ORDER BY a.speaking_url;`

webian:
Another solution to this problem came to my mind but it's dirty:
to sort by `original_url` the `exec_SELECTgetRows` of `getUrlFromCacheBySpeakingUrl()` so the one with cHash (the longer) comes first.

dmitryd:
Sorting by strings will be very slow. On large tables it will take seconds and even indexes will not help.

Classes/Cache/DatabaseCache.php
```diff
@@ -199,6 +199,15 @@ public function getUrlFromCacheBySpeakingUrl($rootPageId, $speakingUrl, $languag
 				'', 'expire'
 		);
 
+		// More than one urldata entry? Then prefer those with cHash
+		if (count($rows) > 1) {
+			if (preg_match('/cHash/', json_encode($rows))) {
+				$rows = array_filter($rows, function ($row) {
+					return strpos($row['original_url'], 'cHash') !== false ? true : false;
```
dmitryd: This ignores all logic below and will break language handling with _DOMAINS. No go.

dmitryd:
The patch will bring back an old bug with _DOMAINS & same URLs for different languages. This solution is no go.

Classes/Cache/DatabaseCache.php
webian: I don't understand, I need one example of 2 original_url that will break with this code. Can you provide?

dmitryd: www.domain.ch/news/contact-us/
www.domain.co.uk/news/contact-us/

URLs are the same but language passed to the function is different because of the _DOMAINS. If you ignore the language, you have 50% chances to find and return the wrong url.

dmitryd: It is better to make sure that there are no duplicates but it is hardly possible after 2.0.15. Now we can only do some more evaluation of `$rowCandidate` to see if cHash is there and use that.

Try to avoid tricks with `json_encode`, etc and make core more obvious.

dmitryd:
Thinking more about it, it is probably wrong to solve the problem on the cache class level because logically this class should not be aware of cHash issues. I have an idea how to solve it in decoder/encoder.

Gregor-Agnes:
Why even cache the urls without a chash? They do not make any sense in TYPO3 (with or without realurl) and thus shouldn't be cached.

dmitryd:
> Why even cache the urls without a chash? They do not make any sense in TYPO3 (with or without realurl) and thus shouldn't be cached.

 I suggest that you google for cHash, what it means and when it must be be used and (more importantly) when it must not be used. Than check your statement again :)

Gregor-Agnes:
Hm, I think getvars for extensions without a chash don't make sense, right? 
Because without chash the request can't get the proper content from the cache table.

If I load eg a news without a chash and without enabled realurl, I get an error (depending on what is configured on chash error). 

That's why I think, urls with ext getvars and without chash shouldn't get cached in the urldata table. They don't exist in TYPO3.

Maybe just two different views. Thx for the great extension though.

dmitryd:
> Hm, I think getvars for extensions without a chash don't make sense, right? 

Wrong.

> Because without chash the request can't get the proper content from the cache table.

Non-cached plugins: [USER_INT](https://docs.typo3.org/typo3cms/TyposcriptReference/ContentObjects/UserAndUserInt/Index.html).

> That's why I think, urls with ext getvars and without chash shouldn't get cached in the urldata table. They don't exist in TYPO3.

:) We all think that something does not exist until we discover it. Just two hundred years ago people did not believe that meteorites existed. Same with URLs without cHash. They exist and perfectly legal.

dmitryd:
@webian I made a different solution, which takes language into account. You can see other improvements in this area [in this branch](https://github.com/dmitryd/typo3-realurl/compare/feature/cHashImprovements).

Gregor-Agnes:
Yeay, right, but we talk about cached content only. 
USER_INTs and meteorits are not subject of that discussion :)

I tried your fix and it seems that the false entries (without chash) won't be created anymore (or they are deleted :) ). 
Everything seems to work perfectly, thx!

dmitryd:
> Yeay, right, but we talk about cached content only. 

No. Realurl encodes and decodes any types of URLs that TYPO3 requires, including URLs for non-cached content. :)

Gregor-Agnes:
I see. But does realurl need to write entries in the urldata table for non-cached content? I recognized, that it only generates entries in that table, if I emptied the FE-Cache first. So I thought: No entry in the FE-Cache should lead to no entry in the urldata. 
Sorry for the noise, but I'm fighting for some days now with the problem and just want to understand. :)

dmitryd:
> I see. But does realurl need to write entries in the urldata table for non-cached content?

Yes. Any `/search/` or `/contact/` url, for example. They still exist on your site and you want them to be resolved fast when visiting.

PR #290, opened 2016-10-05
[TASK] Store crdate when a new record is created in tx_realurl_urldata

webian:
I don't know if there's a particular reason why crdate is not stored since there is the field in the DB.
I think that crdate is very useful when trying to understand what created a particular entry while debugging Realurl configuration. With crdate you can, by example, go back to the HTTP request from the Apache log.


dmitryd:
Thank you!

PR #294, opened 2016-10-10
[BUGFIX] Fix #278: Case sensitive url with BE login

### LINKED ISSUE #278: Case sensitive URLs throw 404 error page
Only with active TYPO3 backend login I get an 404 page while calling an url like "www.example.com/Big-Letters/are-bad.html". Same URL works without active backend login.

I am using realurl 2.0.15 with TYPO3 6.2.27. This issue came up after upgrading from TYPO3 6.2.26 to 6.2.27 and realurl 1.13.x to 2.0.15 at the same time.

lepus:
- Fixes the issue that you will get a 404 when you type in an url with upper case letters, but only if you are logged in to backend

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -966,7 +966,7 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 			$this->throw404('Current URL is invalid');
 		}
 
-		return ltrim($urlPath, '/');
+		return strtolower(ltrim($urlPath, '/'));
 	}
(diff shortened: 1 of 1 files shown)


dmitryd:
I will need some time for checking this. While looks easy, may be other places has to be fixed too.

liayn:
I'm not sure how this is related to being logged in into BE.
Honestly, I dislike this change, since it would make any URL lowercase.

dmitryd:
This should not be a problem because the decoder is not affected by the BE login anymore.

Issue #295, opened 2016-10-11
$requestVariables['L'] should only store INT values to tx_realurl_urldata Table


velletti:
We have a multi langauge setup for our homepage

en.html is mapped to index.php?L=0 
and 
de.html is mapped to index.php?L=1

Rootpage ID = 76 

**Today it happened that the following entry was stored into tx_realurl_urldata :**

 `original_url= L=1%27A%3D0&id=76 
speaking_url = de.html 
request_variables = {"id":"76","L":"1'A=0"}`

and then "de.html" shows the content of the englisch Version. Deleting that row in table 'tx_realurl_urldata', fixes the problem. But it  re-appeared. 

**My (custom made) RealUrl Conf File looks like this:**
`$TYPO3_CONF_VARS['EXTCONF']['realurl'] = array(
    '_DEFAULT' => array(
        'init' => array(
            'enableCHashCache' => 1,
            'appendMissingSlash' => 'ifNotFile',
            'enableUrlDecodeCache' => 1,
            'enableUrlEncodeCache' => 1,
            'emptyUrlReturnValue' => '/',
            'postVarSet_failureMode' => '',
        ),

```
    'cache' => array ( 'banUrlsRegExp' => '/ContactLeadId=|gclid=|type=|(?:^|\?|&)q=/' )

    'redirects' => array(),
    'preVars' => array(
        array(
            'GETvar' => 'L',
            'valueMap' => array(
                // alle sprachen die doch nicht live gehen deaktivieren
                //'en' => 0, //international (needs no url part because its the default language)
                'de' => 1, //germany
                'it' => 2, //italy
                'cz' => 3, //czechrepublic
                'fr' => 4, //france
                'ch_de' => 6, //switzerland - german
                'at' => 7, //austria
                'es' => 18, //spain
                'ch_fr' => 19, //switzerland - french
            ),
            'noMatch' => 'bypass',
        ),
    ),
    'pagePath' => array(
        'type' => 'user',
        'userFunc' => 'EXT:realurl/class.tx_realurl_advanced.php:&tx_realurl_advanced->main',
        'spaceCharacter' => '-',   ..... .... `
```

---

Solution: 

**Classes/Controller/UrlCacheController.php**
Change line 581 -> 583 
`if (isset($requestVariables['L'])) {
    $this->detectedLanguageId =  (int)$requestVariables['L'] ;
}
`

to this:

`if (isset($requestVariables['L'])) {
[...]

PR #296, opened 2016-10-11
#295 - RealUrl will now only store INT values for "L" parameter to tx…

velletti:
…_realurl_urldata Table

before storing/detecting the requestVariables param L= cast the Value to INT to AVOID that params like L=1'A=0   will reach the tx_realurl_urldata table and show wrong language output in Frontend


dmitryd:
Please, fix your TypoScript instead: https://docs.typo3.org/typo3cms/TyposcriptReference/Setup/Config/Index.html#linkvars

You should have `config.linkVars = L(int)` in your TS setup.

It is not realurl's job to change parameters passed by users.

velletti:
Thanks for the Solution that helps me to to stay compatible with the original Version. So i can close my Issue.

PR #298, opened 2016-10-12
fix cHashIncludePageId Error on Root Page

### LINKED ISSUE #297: cHashIncludePageId Error on Root Page
**Issue Description**

We observed the following error message 

> ID parameter needs to be passed for the cHash calculation! As a temporary not recommended
> workaround, you can set $GLOBALS['TYPO3_CONF_VARS']['FE']['cHashIncludePageId'] to false to
> avoid this error."

which is thrown in https://github.com/TYPO3/TYPO3.CMS/blob/TYPO3_7-6/typo3/sysext/frontend/Classes/Page/CacheHashCalculator.php#L131

**Analysis**

This happens when submitting a form with powermail on the root page of a website. The very same form can be submitted on any other page without problems. 

The URL of the root-page call looks as follows:

`https://domain.com/?tx_powermail_pi1%5Baction%5D=create&tx_powermail_pi1%5Bcontroller%5D=Form&cHash=a6fcc3769dcf97303e63c4faf8799fe2`

The speaking-url is correctly generated in 
https://github.com/einpraegsam/powermail/blob/master/Resources/Private/Templates/Form/Form.html#L15
https://github.com/TYPO3/TYPO3.CMS/blob/TYPO3_7-6/typo3/sysext/fluid/Classes/ViewHelpers/FormViewHelper.php#L200

and therefore stored in the database as 

original url: `cHash=a6fcc3769dcf97303e63c4faf8799fe2&id=1&tx_powermail_pi1[action]=create&tx_powermail_pi1[controller]=Form`

speaking url:
`?tx_powermail_pi1%5Baction%5D=create&tx_powermail_pi1%5Bcontroller%5D=Form&cHash=a6fcc3769dcf97303e63c4faf8799fe2`

The problem is, that the recently introduced `$GLOBALS['TYPO3_CONF_VARS']['FE']['cHashIncludePageId']` setting requires the Get-parameter `id` to be set in order to correctly calculate the cHash. This is set correctly when calling any url of the form

`http://domain.com/my/page?my_param=bla`

but not for urls like

`http://domain.com/?my_param=bla` 

**Solution**
The suggested solution is to remove the following line
https://github.com/dmitryd/typo3-realurl/blob/development/Classes/Decoder/UrlDecoder.php#L1212

and thus accept `http://domain.com/?my_param=bla`  as a valid speaking url, which it is, in fact, because otherwise it would be `http://domain.com/index.php?id=1?my_param=bla`.

**Discussion**
This is not yet fully tested and could be discussed for sure, but please, do not reply with comments like https://github.com/dmitryd/typo3-realurl/issues/291#issuecomment-252442787

The introduction of `$GLOBALS['TYPO3_CONF_VARS']['FE']['cHashIncludePageId']` caused quite some problems, especially the fact that realurl was not generating missing cHashes anymore, which would be fine, if it did not store the generated speaking url without a correct cHash in the urlcache table (which is not the case anymore, am I right?).
You are right by saying that correct urls need to be generated in an extension, but in this case the url is generated correctly and works on any page except on the root page.

macjohnny:


macjohnny:
this resolves #297

dmitryd:
https://forge.typo3.org/issues/78021

PR #301, opened 2016-10-17
[FEATURE] add keepParameter option for domains

fvosberg:
if you are using 2 domains for 3 languages, you need to have the L parameter present for the pagepath. Now you have a keepParameter option for the _DOMAINS configuration, to prevent realurl from unsetting the lang param.

Example:

foobar.de/de/
foobar.de/en/
foobar.es

If you configure the _DOMAINS you loose the language param for page path decoding. To prevent this and leave the API backwards compatible, I've introduced a new option for the _DOMAINS section.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -1171,7 +1171,10 @@ class UrlEncoder extends EncodeDecoderBase {
 					// because 'urlPrepend' is typically used for language-based domains.
 					// But 'useConfiguration' can be used to localize postVarSet segment
 					// values. So we change the behavior here comparing to 1.x.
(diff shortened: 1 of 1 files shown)


dmitryd:
> If you configure the _DOMAINS you loose the language param for page path decoding

No, you do not. Path decoding only uses language variable. It is set [here](https://github.com/dmitryd/typo3-realurl/blob/development/Classes/Decoder/UrlDecoder.php#L987) and later used at path decoding.

> To prevent this and leave the API backwards compatible, I've introduced a new option for the _DOMAINS section.

"The API" for realurl does not exist. And realurl always removed that var. So there is nothing about compatibility here.

PR #302, opened 2016-10-18
[TASK] Add svg icons for extension and backend module

jousch:
Since TYPO3 CMS 8.3, extension icons should/could be moved to
Resources/Public/Icons/Extension.svg. I therefore created an svg icon
of your actual "pill icon".

The backend module actually works with TYPO3 CMS 7.6 and above.
An svg version of the icon could be added therefore. I suggested to slightly
modify it by adding a background and removing the drop shadow.


jousch:
image

image

faulancr:
i would suggest to reduce the icon to a more reduced version. 
[realurl_icon.svg.zip](https://github.com/dmitryd/typo3-realurl/files/545689/realurl_icon.svg.zip)
bildschirmfoto 2016-10-22 um 10 49 54

jousch:
@faulancr They look nice! I like v2.

@dmitryd, what do you think about? I also can imagine an icon with a text in the icon like "/url/".

dmitryd:
@jousch Wow, that pill looks cool :) While @faulancr icons looks more in TYPO3 style, I still like the pill more. Let me think. One of these definitely should be integrated. Or both. May be, pill for one release and that link icon since the second release from now. I really want to see that cool pill first :)

faulancr:
@dmitryd the go with version 2 ;) The Linkchain looks quite similar to the pill.

dmitryd:
realurl_icon.svg.zip looks more in TYPO3 style. What do you all think?

jeremieroulin:
I vote for realurl_icon_3.svg.zip! :)

faulancr:
the reason i put a speaking bubble into it, is that the chain is more a symbol for "link" than for "url", which in the end is quite similar but not the same.

dmitryd:
Hm. Let's make a vote through twitter :)

https://twitter.com/dmitryd/status/792431489459838978

dmitryd:
I could not leave those two icons. So I added them too and made possible to select them for the module :)

PR #305, opened 2016-10-19
add 2nd index to tx_realurl_uniqalias_cache_map

thegass:
add index on url_cache_id  to tx_realurl_uniqalias_cache_map to speedup deletes.
the combined index check_existence is not used in queries like
DELETE FROM tx_realurl_uniqalias_cache_map WHERE url_cache_id = XXXX .


dmitryd:
This index may prevent usage of the first index when doing main queries. I do not see advantages.

dmitryd:
Rejected for now.

Issue #318, opened 2016-11-02
fixedPostVars need to be fully matched


liayn:
If a fixedPostVarSet is only partially matched on decoding, the input is currently accepted and the requestVariables are created. This is in our opinion a big issue, since this will cause duplicate content problems as two urls will point to the same content.

Example:
A fixed postvarset is configured to have news records shown as `/year/title`, like `example.com/2015/news-title`. The root page is configured to display the news detail view if a news is provided via GET.
Calling the page `example.com/` alone delivers the homepage with a 200 status code.
Calling the page `example.com/asdf/` delivers the exact same output as above, also with a 200 status code.
This is a no-go.

`\DmitryDulepov\Realurl\Decoder\UrlDecoder::decodeFixedPostVars` contains a
```
// TODO Is it correct to break here? fixedPostVars should all present!
break;
```

I would strongly suggest that this is **not** correct here.
All variables have to be set of a fixedPostVarSet, otherwise no other logic has a chance to detect the malformed URL.

The pull request would be really easy ;-)

dmitryd:
The comment you mentioned is my user/author dilemma. Realurl 2.x is compatible with 1.x in this behavior, so I cannot really break it. As a user I fully agree with you but as the author I cannot break other existing sites that rely on this functionality. And I know that such sites exist.

Generally you should design your URL structure in a way that does not allow such issues. Realurl cannot be too clever because it is a hack on top of TYPO3. The simpler your URLs are, the better. By "simpler" I mean that they should not be too clever. Ideally you should have a dedicated page for news single view. I know this is not always desired but than realurl is also very complex and it cannot solve absolutely every problem and use case.

I do not think this can be enabled **by default**. However, this can be done with a configuration option. Something like `init/fullFixedPostVarEvaluation` or a better name.

dmitryd:
Example of a fixedPostVar with variable number of parameters:

/path/to/detail/page/my-info-item-title/
/path/to/detail/page/my-info-item-title/mark-unread/

The first one is a `lookUpMap`, the second one is a controller action with `noMatch=bypass` and `mark-unread` equals to `markUnread` action. That one is used in a project I know.

liayn:
Yes, an option to enable this check would be totally enough for me. I would even make that option per set and not globally.
I'll prepare a PR for that.

liayn:
Regarding the example above. If a `bypass` is used, this is still a full evaluation IMO.

dmitryd:
[...]

PR #319, opened 2016-11-02
[FEATURE] Add option to enforce full evaluation of fixedPostVarSets

liayn:
Fix: #318

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -400,9 +400,11 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 
 			$previousValue = '';
 			foreach ($postVars as $postVarConfiguration) {
+				if (!is_array($postVarConfiguration)) {
+					continue;
+				}
 				$this->decodeSingleVariable($postVarConfiguration, $pathSegments, $requestVariables, $previousValue);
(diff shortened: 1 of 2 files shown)


liayn:
@dmitryd ping ;-)

dmitryd:
I already reviewed the code and it looks ok to me but I need to test this :)

liayn:
Did you find time already?

dmitryd:
No. I work on realurl either in my personal time, or when I have some time from my employer. Currently none had any free time.

Even if I merge this right now, you will still need to wait for the release, which will need more testing. So it does not change much if I merge it now or in a week or in two weeks.

Classes/Decoder/UrlDecoder.php
```diff
@@ -400,9 +400,11 @@ protected function decodeFixedPostVars($pageId, array &$pathSegments) {
 
 			$previousValue = '';
 			foreach ($postVars as $postVarConfiguration) {
+				if (!is_array($postVarConfiguration)) {
```
dmitryd: This should not be needed because there is a call to `getConfigurationForPostVars`, which resolved postVars to arrays.

dmitryd: Ah, no, I am wrong here...

PR #324, opened 2016-11-09
added option to deactivate urlcache on single pages

stiXits:
In situations where a misconfigured system or plugins excessively using get parameters may pollute the url cache (even to a failing state), deactivating the url cache for single pages may be a work around or even a solution (for example large formhandler formulars).
These additions put the "deactivate realurl cache" option back into the pages' properties section and respect this option on creating a new url cache entry.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -1276,7 +1276,16 @@ class UrlEncoder extends EncodeDecoderBase {
 	 * @return void
 	 */
 	protected function storeInUrlCache() {
-		if ($this->canCacheUrl($this->originalUrl)) {
+
+                $cacheDeactivated = $this->databaseConnection->exec_SELECTgetRows(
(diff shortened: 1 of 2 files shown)


dmitryd:
Firsts, this is worse than using [`banUrlsRegExp`](https://github.com/dmitryd/typo3-realurl/wiki/Configuration-reference#banurlsregexp) because can be easily abused.

Secondly, "cache" remains only in class names. It is not cache anymore. It is data. Realurl 2.x relies heavily on having this data. It is especially important when decoding.

So I do not think this option is needed. Use the configuration or re-configure the system. The option you try to bring back, will not cure the illness, it will mask symptoms while introducing ways for various people to make realurl work worse.

I hope I describe it clear enough...

dmitryd:
Based on my experience, people WILL abuse this setting and than come asking for help. So use `banUlrsRegexp` instead.

PR #330, opened 2016-11-16
[BUGFIX] MountPoint with Path CacheEntry

cdfre:
Additional GET-Parameters resulted in a 404 Error page


dmitryd:
Sorry but this needs a proper bug report and explanation. It is no go to post a change without proper explanation unless it is a typo or a trivial fix.

cdfre:
I'll try to explain the problem with an example.

- Folder A (uid: 1)
-- Page A (uid: 2)
- Root Page (uid: 3)
-- Subpage (uid: 4)
-- Mount Point (=> Folder A) (uid: 5)

The Path to Page A would be:
www.webpage.com/mount-point/page-a/

This URL works and the page can be opened in the browser.
Internally the path would be translated to ?id=2&MP=1-5

But wenn i add an additional parameter to the url:
www.webpage.com/mount-point/page-a/?mytestparameter=1

The http requests results in a 404 error.
In RealURL the MP-Parameter is missing from the resolved cache entry.

dmitryd:
Thank you. I will check this use case.

corth:
I can confirm that. The patch from @cdfre works for me!

cdfre:
Can you give me a status or more feedback regarding this pull request? @dmitryd

adrienjacob:
I also confirm I had the problem, and it got solved with the suggested commit. Thanks @cdfre

dmitryd:
@cdfre I appreciate your effort and work. But this is a free project for me, which I usually do after 8+ hours of the office work. So I'll check whenever I have time for it. Sorry :(

dmitryd:
Thank you! I am sorry that it took so long but I finally found spare time to review it :)

PR #331, opened 2016-11-16
[BUGFIX] ensure removeIgnoredParametersFromQueryString returns string

maritwho:
If the url has no query string, removeIgnoredParametersFromURL sets
the variable to null. This causes, that the function
removeIgnoredParametersFromQueryString returns null as well. But if
it returns null instead of empty string, the
removeIgnoredParametersFromURL function appends an senseless
question mark.


dmitryd:
Thank you.

PR #336, opened 2016-11-22
[BUGFIX] fix #335: UrlEncoder now recognises default language for empty L parameter

### LINKED ISSUE #335: Translated URL Paths in Default Language
I am actually not sure if this is a real issue, but I've had some trouble which seems related to this behaviour:

Assume I have a page called "Trouble Page" (ID: 123). It's German (language ID: 1) translation is "Problem-Seite".
When I view the default and the German version (not logged into the backend) I find all cache table entries I am expecting afterwards, but actually I get an additional entry.

Having a look at `tx_realurl_urldata` I see the following entries:
```
original_url | speaking_url
============================
id=123       | /trouble-page/
L=1&id=123   | /de/problem-seite/
L=&id=123    | /problem-seite/
```

The third entry is bugging me. It has the translated path segments but no language prefix.

As far as I can tell, this is not a problem in general since the first entry is used for generating links in the default language, but I did find some cases (though I do not know how to reproduce them exactly) in which this third entry showed up (on links generated in the default language). Some of those cases led to a 404 error, some were working (showing the default language version of the page).

Can you tell me, if this third kind of entry is generated on purpose?

/edit

Now I found a way to reproduce the problem in the front end using a bit of TypoScript. It's actually dead-simple.

```
10 = TEXT
10.typolink.parameter = 123
20 < .10
20.typolink.additionalParams = &L=1
30 < .10
30.typolink.additionalParams = &L=
```

The first link will point to the default version with the normal URL. The second will point to the translated version. The third will also point to the default version but it will be using the translated path segments.

s-ude:


dmitryd:
[Misconfiguration](https://github.com/dmitryd/typo3-realurl/wiki/Notes-for-Integrators#configuring-languages), no need for the change.

PR #337, opened 2016-11-22
[BUGFIX] Dynamic Asset Path for Backend Module

marcorieser:
We install the extension with composer into the typo3/ext folder. Because the path to the assets were hard coded (typo3conf/ext) the assets could not be loaded. With this fix, the path gets generated correct irrespective of the extension installation path.


dmitryd:
Thank you.

PR #339, opened 2016-11-25
[TASK] reduce amount of calls to limit entries in database tables

franzkugelmann:
on bigger websites (in our case ~50.000 pages and ~1.000 editors) the sql-statement to limit the number of entries in i.e. pathdata is called every few seconds. Because % 5 means around every fifth's time a new entry is written into that table. From my point of view it would be fitting to do this far more rarely.
It helped in our case to considerably improve DB performance.
% 5000 ist just a suggestion we use currently. Feel free to modify according to your opinion on that.


thegass:
+1 the default value kills performance on big sites.

dmitryd:
Thank you!

PR #340, opened 2016-11-25
[TASK] Replace underscore with hyphen while saving explicit pathdata in backend

franzkugelmann:
this is meant as hint to a behaviour which is confusing our editors currently. If the fill the pathdata-field explicitely they expect the url to be exactly like what they entered. Since in former realurl-version they used the underscore quite often, they still enter underscores. The underscore stays there if they save the page, but in frontend realurl creates a hyphen instead.
We are totally fine with the hyphens but suggest to replace underscores with hpyhens already while saving to avoid confusion.


dmitryd:
Thank you!

PR #347, opened 2016-12-01
[BUGFIX] fix urldata cache entries

maritwho:
Store the url in urldata cache in such way
to be found with $this-speakingUri in decode
process.

* add leading slash
* do urldecode

The cache entries are not found because of these two points above.


dmitryd:
This is not going to work because URLs in the cache are searched without leading slash. The problem with url encoding is solved at https://github.com/dmitryd/typo3-realurl/commit/03906d1d9a4a639f4bb73640786baf8ea4fce14c

maritwho:
Ok great,

thank you for solving :-)

PR #352, opened 2016-12-06
[BUGFIX] use mb_substr for alias

franzkugelmann:
Since substr ist not multi-byte save, realurl cannot create an alias if you have for example a news title with a german umlaut at exactly the position where you want to cut the string. Then the alias ist just empty.
Using mb_substr fixes that issue.


dmitryd:
I cannot merge this because:
1. `mbstring` is not the default PHP extension. It is not always installed.
1. TYPO3 provides functions for that which we can use.

I made a different solution.

PR #359, opened 2016-12-14
[BUGFIX] Cover all cases for appendMissingSlash

iamandrewluca:
Add `ifNotFile` in the regex to also match that option

I have a Multi language (`DE, EN, ES`) TYPO3 installation with realurl installed and configured. Default language `DE`
When I request `domain.com/es/` I get the Spanish version
But when I request `domain.com/es` without slash at end, I get German version.

This fix solves my problem.

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -234,8 +234,8 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 			if ($this->appendedSlash && count($options) > 0) {
 				foreach ($options as $option) {
 					$matches = array();
-					if (preg_match('/^ifNotFile|redirect(\[(30[1237])\])?$/', $option, $matches)) {
-						$code = count($matches) > 1 ? $matches[2] : 301;
+					if (preg_match('/^(ifNotFile|redirect(\[(30[1237])\])?)$/', $option, $matches)) {
+						$code = count($matches) > 2 ? $matches[3] : 301;
 						$status = 'HTTP/1.1 ' . $code . ' TYPO3 RealURL redirect';
 
 						// Check path segment to be relative for the current site.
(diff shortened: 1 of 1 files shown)


dmitryd:
`ifNotFile` is already handled above in the file. I could not reproduce the issue with this configuration:

```php
$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl'] = array (
	'test.dev' => array(
		'init' => array(
			'appendMissingSlash' => 'ifNotFile,redirect[302]',
		),
...
```

iamandrewluca:
> ifNotFile is already handled above in the file.
Here ?
https://github.com/iamandrewluca/typo3-realurl/blob/80f58dc1583aeb7cb2cd525e1016ba1a4d93fcf2/Classes/Decoder/UrlDecoder.php#L225

dmitryd:
@iamandrewluca 

> However it seems at this line ifNotFile can be in $options and is not handled in if (preg_match... and redirect is not done

`ifNotFile` only restricts how redirects are done. It does not have to be in the regex to work.

iamandrewluca:
Well I think now I understand, I should set `appendMissingSlash` option to `redirect,ifNotFile` ?

dmitryd:
That's right. This thing is historical since realurl 1.x. To me it does not look like we need `ifNotFile` at all...

iamandrewluca:
Ok, thanks @dmitryd !
> To me it does not look like we need ifNotFile at all...

What if we have a file?

I set my option to `ifNotFile,redirect` everything works fine, except my
```
$GLOBALS['TYPO3_CONF_VARS']['FE']['pageNotFound_handling'] = '/404';
```
It gets into a infinite redirect. I set it to `/404/` and I get another error which is whole another problem
Thanks again!  
For now I will close this issue.

dmitryd:
Don't use speaking url there. Use normal `/index.php?id=12345`.

Issue #378, opened 2017-01-20
Wrong redirects for UTF8 URLs after changing configuration


mvoehringer:
Situation:
We want to change an existing Site that uses utf8 URLs for Chines to an english URL schema. Thats why we set the "languageExceptionUids" option for Chines. After clearing the some caches the new schema works as expected. 

Problem:
When users try to access the expired Chines version of the URLs, Realurl tries to redirect them to a crazy version like this: domain.com/some-english-segments%8C服务和支持/现场支持.

Analyses: 
The problems is in the function  UrlDecoder->decodePath Line 515 here RealUrl compares and URL encoded Version and unencode version: 

this->speakingUri: 
`cn/%E6%9C%8D%E5%8A%A1%E5%92%8C%E6%94%AF%E6%8C%81/%E7%8E%B0%E5%9C%BA%E6%94%AF%E6%8C%81/`

this->expiredPath: 
`服务和支持/现场支持`

Please see my change request about an simple way to fix this. 

* Realurl Version 2.1.5
* TYPO3 6.2

adrienjacob:
Indeed, as I said in another thread, I believe none of the pages in CJK languages are finding cache entries properly. I added this on line 157 of UrlDecoder.php (before calling runDecoding) :
$this->speakingUri = urldecode($this->speakingUri);
Maybe your pull request is "cleaner" ?

dmitryd:
In my current development version paths in the cache are not encoded at all:

![url-cache](https://cloud.githubusercontent.com/assets/306133/22787626/05fb287a-eeee-11e6-969a-5a0d570238f4.png)

Could you post a picture of yours?

dmitryd:
-- removed my previous comment because it is not relevant --

PR #379, opened 2017-01-20
[BUGFIX] Wrong redirects for UTF8 URLs after changing configuration #378

mvoehringer:
This is only an example that fixes my issue, to solve the issue speakingUri should be urldecoded global.

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -507,11 +507,12 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 			$remainingPathSegments = $savedRemainingPathSegments;
 		}
 		if ($result && $this->expiredPath) {
-			$startPosition = (int)strpos($this->speakingUri, $this->expiredPath);
+			$speakingUri = urldecode($this->speakingUri);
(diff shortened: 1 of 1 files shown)


dmitryd:
The problem here is that cache entries should be url encoded for the path part but they were not. This is fixed in https://github.com/dmitryd/typo3-realurl/commit/03906d1d9a4a639f4bb73640786baf8ea4fce14c now.

dmitryd:
Unfortunately it makes old entries invalid and useless now :(

mvoehringer:
Good news, that you fixed this problem. Loosing all the old urls is not a good news, what is the problem with these old urls? Is the problem, that all the urls are encoded in the cache after you update (like it was in older realurl versions)?

dmitryd:
@mvoehringer Yes, they are not encoded in the cache. I am currently thinking of making an update procedure for such entries but I need to find reliable way to detect them. And it can timeout because it can take a lot of time if there are many entries.

The best way is to clear FE cache. Pages will be regenerated as well as urls.

dmitryd:
See also https://github.com/dmitryd/typo3-realurl/commit/ef79bf76197561ce73d5e1559926938a45fd5cfa

dmitryd:
@mvoehringer The update wizard is integrated now.

Issue #357, opened 2016-12-13
Speaking path with override not decoded properly?


adrienjacob:
Hi,

When setting the "speaking path" parameter with override, I noticed the URL would not be decoded properly, unless the page is actually displayed in a menu (and thus cached), or is a direct child of the root page.
Maybe I'm wrong, but I believe there's a problem in UrlDecoder.php with function getRootPageIdForPage (used by searchForPathOverrideInPages and searchForPathOverrideInPagesLanguageOverlay). This function is supposed to return the root page ID of the current page, but instead returns the topmost root page. Here's my setup : 
Root (not set as "is root")
-> Domain 1 (with "is root" set, domain record and root id configured in realurl_conf.php)
-->Page 1
--->Page 2
--> Page 3
-> Domain 2 (same)
etc.
getRootPageIdForPage(Page 2) returns "Root" instead of "Domain 1".

This fixed my problem :
```
	protected function getRootPageIdForPage($pageUid) {
		$rootLineUtility = GeneralUtility::makeInstance('TYPO3\\CMS\\Core\\Utility\\RootlineUtility', $pageUid);
		/** @var \TYPO3\CMS\Core\Utility\RootlineUtility $rootLineUtility */
		$rootLine = $rootLineUtility->get();
		
		// START CHANGE
		$result = 0;
		if (is_array($rootLine) && count($rootLine) > 0) {
			$index = count($rootLine)-1;
			while($index) {
				if ($rootLine[$index]['is_siteroot']) {
					$result=(int)$rootLine[$index]['uid'];
					break;
				}
				$index--;
			}
		}
		
		return $result;
		// END CHANGE
	}
```

dmitryd:
Looks nice, the only thing I'd like to change is the loop direction. Starting from zero is most likely will reach the root page faster. It does not make much difference though.

I will have a look at it.

adrienjacob:
Ok thanks... Otherwise I found https://github.com/hannesbochmann/typo3-realurl/commit/0cad20b4977be6d2664e406b4d1f3da8ff32e9d4 solves the same problem differently, maybe in a better way.

dmitryd:
Yes, I merged that, so the issue can be closed :) Thank you.

Related #259.

Issue #348, opened 2016-12-02
respectSimulateStaticURLs & pages with aliases


adrienjacob:
Hello,

I'm in the process of moving from 	simulatestatic to realurl, and I'd like to keep backwards compatibility to avoid broken external links. I used the respectSimulateStaticURLs parameter, and it works ok... except for pages with aliases. I just tried with realurl v1.13.7 and it works, so it may be a regression in v2.x ?

Thanks in advance for your help !

adrienjacob:
I believe I found the culprit in UrlDecoder.php, line 1169 : 
Change the regex
`(!$this->configuration->get('init/respectSimulateStaticURLs') || !preg_match('/^[a-z0-9\-]+\.(\d+)(\.\d+)?\.html/i', $this->siteScript))`
To
`(!$this->configuration->get('init/respectSimulateStaticURLs') || !preg_match('/^[a-z0-9\-]+\.(.+)(\.\d+)?\.html/i', $this->siteScript))`

PR #388, opened 2017-01-31
[BUGFIX] Fix decoding of page with override & page aliases with respectSimulateStaticURLs

adrienjacob:
Cf issue https://github.com/dmitryd/typo3-realurl/issues/357


adrienjacob:
Added second commit on another issue, explained in https://github.com/dmitryd/typo3-realurl/issues/348.

adrienjacob:
Resubmitting in two separate PR, as the two issues are not related...

Issue #367, opened 2017-01-05
Cache not updated when changing "override" or "exclude from speaking URL"


adrienjacob:
When changing the value of "Exclude from speaking URL" or "Override the whole page path", the cache is not refreshed. I fixed it by changing line 182 of Hooks/DataHandler (in function expireCache)

```
if (isset($databaseData['hidden']) || isset($databaseData['tx_realurl_pathoverride']) || isset($databaseData['tx_realurl_exclude'])) {
		$expireCache = TRUE;
}
```

dmitryd:
See #392

PR #392, opened 2017-02-03
[BUGFIX] Cache not updated when changing "override" or "exclude from speaking URL"

adrienjacob:
Cf. https://github.com/dmitryd/typo3-realurl/issues/367


dmitryd:
Thank you.

PR #393, opened 2017-02-03
Add palette options for mount points

adrienjacob:
This adds palette options for mount points in the back-end ("exclude from", "speaking url", "override"...)


dmitryd:
Thank you!

Issue #395, opened 2017-02-05
Clarify url encoding in cache for CJK languages


adrienjacob:
I open a new ticket, partly related to issues https://github.com/dmitryd/typo3-realurl/issues/310 and https://github.com/dmitryd/typo3-realurl/issues/378 and PR https://github.com/dmitryd/typo3-realurl/pull/379, as I believe all are linked to a wider issue.

After spending some time trying to solve another problem (the links to the detailed view of tt_news were not working properly with CJK languages - the postvarset was all % encoded, and the URL path segments were not, resulting in bugs in some browsers), I got somewhat confused on the intended behaviour of the urldata : is it supposed to be stored with rawurlencode() or not ? Because there are conflicting behaviours between
- UrlEncoder/createPathComponentUsingRootline : this manages the path segment, and is not rawurlencoded
- UrlEncoder/encodeUrlParameterBlockUsing* : these functions seem to manage postvar segments, and are rawurlencoded

I temporarily fixed all this with https://github.com/adrienjacob/typo3-realurl/commit/f3403218366793e030613bb322519c22c0298aec. But that's assuming the cache is supposed to store non rawurlencoded data, and I'm not sure of that. Besides, it doesn't make much sense to rawurlencode postvars in the first place, to decode them later on, so it's clearly not a satisfying solution in the long run.

adrienjacob:
Thanks @dmitryd !

PR #396, opened 2017-02-06
[BUGFIX] rawurlencode query string parameters in tx_realurl_urldata

bnf:
Currently `speaking_url` and `original_url` are stored unencoded in
tx_realurl_urldata. When looking up a cache entry for
the current request $TSFE->siteScript is effectively compared with
`speaking_url`. The former is url encoded, the latter not.
Thus a cache entry will never be found if the url contains characters
that need encoding (like square bracket).

This will break in the following case:
If no cache entry was found, a temporary cacheEntry was created which
did not blow up unless one used fixedPostVars, additionalGetParams (e.g GETvars
not mentioned in fixedPostVars and thus appended to the url), and
noMatch=bypass (to remove some GETvars, e.g tx_foo_bar[action])
and a cHash.
realurl tries to recreate the parameters as temporary cacheEntry
from the url, but it doesn't know the GETvar value used for the
noMatch=bypass rule, so that GETvar will be missing and the cHash
comparision will fail.

### DIFF
--- a/Classes/EncodeDecoderBase.php
+++ b/Classes/EncodeDecoderBase.php
@@ -98,10 +98,11 @@ abstract class EncodeDecoderBase {
 	 * parameters.
 	 *
 	 * @param array $parameters
+	 * @param bool  $rawurlencodeParamName
 	 * @return mixed
 	 */
-	protected function createQueryStringFromParameters(array $parameters) {
-		return substr(GeneralUtility::implodeArrayForUrl('', $parameters), 1);
+	protected function createQueryStringFromParameters(array $parameters, $rawurlencodeParamName = true) {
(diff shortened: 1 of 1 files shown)


adrienjacob:
Thanks, this addresses the issue mentioned in other tickets, such as https://github.com/dmitryd/typo3-realurl/issues/395.
I was wondering whether "speaking_url" is really meant to be urlencoded indeed, and arbitrarily decided it would be "no", so my commit does the opposite of your solution (decodes what is encoded), as does this other PR https://github.com/dmitryd/typo3-realurl/pull/379.
We should just decide once and for all, and be consistent throughout the code.

dmitryd:
Actually it should be encoded because it should be used as is when obtained from the cache.

I believe this is resolved by https://github.com/dmitryd/typo3-realurl/commit/03906d1d9a4a639f4bb73640786baf8ea4fce14c. Please, comment if not.

PR #399, opened 2017-02-10
Crossdomain links when rootpage_id is not defined

adrienjacob:
The documentation says we are not forced to set the rootpage_id in
realurl_conf.php, but can use instead the domain records in typo3 as an
alternative.

However, cross domain links (with typolinkEnableLinksAcrossDomains
enabled) currently work only if each domain has been configured in
realurl_conf.php, with their rootpage_id defined. If using a [_DEFAULT]
config for all domains, cross domain links do not work properly as the
hostname cannot be determined by the setHostnamesForEncoding function.
And then, setRootPageIdFromDomainRecord tries to locate the sys_domain
record, but using the current hostname, not the targeted domain.

### DIFF
--- a/Classes/Configuration/ConfigurationReader.php
+++ b/Classes/Configuration/ConfigurationReader.php
@@ -435,6 +435,9 @@ class ConfigurationReader {
 					}
 				}
 			}
+			if (empty($this->hostName)) {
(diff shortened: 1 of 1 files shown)


dmitryd:
Thank you a lot!

PR #401, opened 2017-02-11
[BUGFIX] Page aliases with respectSimulateStaticURLs

adrienjacob:
Cf. https://github.com/dmitryd/typo3-realurl/issues/348


dmitryd:
Looks good but I plan to update the changed part of regexp to be less wide. Something like `[a-z0-9_\-]` to follow definition of alias exactly as it is in `$TCA`.

adrienjacob:
Thanks !

PR #402, opened 2017-02-11
[BUGFIX] #357 : decoding of page with override

adrienjacob:


adrienjacob:
Actually, I'm not sure we had exactly the same issue as in https://github.com/dmitryd/typo3-realurl/pull/261, but the suggested fix looks cleaner, and may do the job for our problem... I'll test it when I get time.

adrienjacob:
I confirm https://github.com/dmitryd/typo3-realurl/pull/261 works, and answers my problem, so I close this one.

PR #408, opened 2017-02-16
[FEATURE] Disable realurl encoding with TSFE register

astehlik:
Encoding of URLs can now be temporarily disabled by setting
a boolean true value to the tx_realurl_disable register.

This can be used for creating a permalink for example:

```
10 = COA
10 {
	10 = LOAD_REGISTER
	10.tx_realurl_disable = 1

	20 = TEXT
	20.value = Permalink
	20.typolink {
		parameter.field = uid
		forceAbsoluteUrl = 1
	}

	30 = RESTORE_REGISTER
}
```

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -1105,7 +1105,8 @@ class UrlEncoder extends EncodeDecoderBase {
 	 * @return bool
 	 */
 	protected function isRealURLEnabled() {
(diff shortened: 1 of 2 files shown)


dmitryd:
Makes sense except that I am not very happy with having both `tx_realurl_disable` and `tx_realurl_enable`. People will start confusing them. Let me think.

dmitryd:
I renamed the register to avoid confusion.

Thanks for the idea!

PR #410, opened 2017-02-17
fix typo (linksVar --> linkVar)

talleux:


talleux:
The typo is in error message, not in the code...

talleux:
There is the same typo in the wiki page. I don't know how to change it there:
https://github.com/dmitryd/typo3-realurl/wiki/Installing-and-enabling-the-extension

dmitryd:
Not needed any more because the message has changed.

PR #416, opened 2017-02-22
[BUGFIX] Fixes fetching of aliases from the cache as well as un-expiring specifically requested aliases - fix #415

### LINKED ISSUE #415: Aliases not getting updated on changes in tx_news
I noticed that in my TYPO3 installation, when I update the Header of a simple page and refresh the frontend, the newly generated URL is displayed on links to that page, which is what I expect.

Now when I change the title of a news record from the tx_news extension, the change in the title does not get reflected in the URL in the frontend, which is not what I expect.

Taking a look into the code I found the problem to be the `getFromAliasCache` (line 967) method of the `UrlEncoder`. There, the comment says

    // We use 'expire=0' condition only if no specific alias is requested. If a specific alias
    // is requested, we also can fetched expired aliases and un-expire them.

which is the desired behaviour. When a specific alias is requested, it's needed and thus should be un-expired. However, what the code does is essentially the exact opposite. Looking a few lines further, on line 978 I found the last addition to the `$where` of the select query to be:

    ($onlyThisAlias ? ' AND value_alias=' . $this->databaseConnection->fullQuoteStr($onlyThisAlias, 'tx_realurl_uniqalias') . ' AND expire=0' : '')

which boils down to "If a specific alias is required, we fetch it if it is not expired. If no specific alias is requested, we fetch everything". This results in the above mentioned problem that the extension fetches expired aliases for a given id and uses the first alias value found, which will always be the first alias to exist for that record.

Moreover, this also results in line 985, which would theoretically un-expire an alias to never be executed, since the check is

    if ($onlyThisAlias && $row['expire'] >0)

while the query only returns rows with expire set to 0 if `$onlyThisAlias` is set. The solution is simply to just fix the query call by moving the `' AND expire=0` part to the exception part of the ternary expression, making it look like this:

    ($onlyThisAlias ? ' AND value_alias=' . $this->databaseConnection->fullQuoteStr($onlyThisAlias, 'tx_realurl_uniqalias') : ' AND expire=0')

I have tried this with my current installation and suddenly everything works like a charm and changes to news titles are reflected immediately. I hope my explanation of the issue was not too confusing. A PR is underway.

TheKvist:
Fixes the issue mentioned in #415 by correcting the check to build the query

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -975,7 +975,7 @@ class UrlEncoder extends EncodeDecoderBase {
 				' AND field_id=' . $this->databaseConnection->fullQuoteStr($configuration['id_field'], 'tx_realurl_uniqalias') .
 				' AND tablename=' . $this->databaseConnection->fullQuoteStr($configuration['table'], 'tx_realurl_uniqalias') .
 				' AND lang=' . intval($languageUid) .
-				($onlyThisAlias ? ' AND value_alias=' . $this->databaseConnection->fullQuoteStr($onlyThisAlias, 'tx_realurl_uniqalias') . ' AND expire=0' : ''),
+				($onlyThisAlias ? ' AND value_alias=' . $this->databaseConnection->fullQuoteStr($onlyThisAlias, 'tx_realurl_uniqalias') : ' AND expire=0'),
 			'', 'expire'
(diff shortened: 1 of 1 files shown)


Zockman:
Just tested and it worked.

dmitryd:
Thank you!

PR #425, opened 2017-03-03
Do not force $this->tsfe->absRefPrefix in postProcessEncodedUrl

geldmacher:
Perhaps i do not get every possible scenario here, but in one scenario this method does not work correctly.

If you have www.domain.de (german) and de.domain.com (english) in one page tree and www.domain.ch (german) and ch.domain.com (english) in another page tree, it is not possible to link from de.domain.com to ch.domain.com. The Link on de.domain.com will always be www.domain.ch, if www.domain.ch is the first domain record in the page tree. 

Because of $this->tsfe->absRefPrefix the urlPrependRegister never gets used an so it can never set the right domain.

Not using absRefPrefix at all is no solution, because it is needed in my scenario. 

As i understand it, the urlPrependRegister is always right, so why bypass it with $this->tsfe->absRefPrefix?
If we just delete it form the url, the urlPrependRegister works as aspected.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -147,7 +147,7 @@ class UrlEncoder extends EncodeDecoderBase {
 			// current host always. See http://bugs.typo3.org/view.php?id=18200
 			$testUrl = $parameters['finalTagParts']['url'];
 			if (preg_match('/^https?:\/\/[^\/]+\//', $testUrl)) {
-				$testUrl = preg_replace('/https?:\/\/[^\/]+\/(.*)$/', $this->tsfe->absRefPrefix . '\1', $testUrl);
+				$testUrl = preg_replace('/https?:\/\/[^\/]+\/(.*)$/', '\1', $testUrl);
 			}
 
 			list($testUrl, $section) = GeneralUtility::revExplode('#', $testUrl, 2);


dmitryd:
This will completely remove `absRefPrefix` from URLs, which is not acceptable. I do not really understand why it does not work in your case but I think it is related to your setup. So it is not a reason to break existing behavior for everybody else.

What is the value of `absRefPrefix` that you use? You should have only `/` there with realurl.

geldmacher:
In my TYPO3 setup i have enabled `absRefPrefix` with `config.absRefPrefix = http://www.domain.tld/`, so every link is absolute by default.

In this case `$parameters['finalTagParts']['url']` is always absolute. If you are linking to another domain (pagetree) the domain in this absolute url is always the first domain record of this pagetree. So the resulting absolute link is wrong in my case if i am linking to the english content.

Second problem is, that `$this->tsfe->absRefPrefix` is wrong in this case, because it is the domain of the current pagetree and not the domain of the pagetree i am linking to. 

I can solve this Problem with the `$urlPrependRegister` because it always has the correct domain if the encode and decode section in `$TYPO3_CONF_VARS['EXTCONF']['realurl']['_DOMAINS']` is used. But because of having relative url's as array keys it is never used in my case.

I hope it is a bit clearer now!? It is hard to explain. =D

dmitryd:
It is clear but I can't apply your change anyway because all urls will become relative. If `config.baseURL` is absent, all links will stop working.

Use `config.absRefPrefix = /` or do a customs solution.

dmitryd:
The problem you are describing is not new to realurl. When the feature of linking to another domain was introduced, there was a [disclaimer](https://docs.typo3.org/typo3cms/TyposcriptReference/Setup/Config/Index.html#typolinkenablelinksacrossdomains) made:

> Disclaimer: it must be understood that while link is generated to another domain, it is still generated in the context of current domain. No side effects are known at the time of writing of this documentation but they may exist. If any side effects are found, this documentation will be updated to include them.

So you found a first side effect years after the feature was added :)

dmitryd:
On the other hand, `typolink` function skips `absRefPrefix` if link goes to another domain. SO it makes sense for realurl to do the same. But than it will be a different change than you propose.

geldmacher:
I understand. As i said "Perhaps i do not get every possible scenario here". =) It was a "quick shot" to have something to talk about. Thanks for your detailed feedback!

I currently solved it by overwriteing 
`$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['tslib/class.tslib_content.php']['typoLink_PostProc']['realurl'] = 'DmitryDulepov\\Realurl\\Encoder\\UrlEncoder->postProcessEncodedUrl';` with my own method.

Do you think this is a stable aproach? Also with regard to future RealUrl and TYPO3 updates?

Perhaps my use-case is too unique for a switch with a configuration parameter!? What do you think?

dmitryd:
I think I will replicate the behavior of the `typolink`: if the root page id of the encoded url is not the same as the current root page id and `absRefPrefix` includes domain name, there should be no `absRefPrefix`. But I need to test this idea.

geldmacher:
Let me now if you need help with this.

dmitryd:
Found that urlPrepend should also be considered (language domains) and updated the PR.

geldmacher:
I will update my System to this new version soon and will give you some feedback. The code looks very promising! Thanks! :D

Issue #199, opened 2016-07-18
Add option to configure pageTitleFields


Kitzberger:
We'd like to alter the set of pageTitleFields (\DmitryDulepov\Realurl\EncodeDecoderBase::$pageTitleFields) that is currently hardcoded to `array('tx_realurl_pathsegment', 'alias', 'nav_title', 'title', 'uid')`.

With realurl 1.x it was possible to change those fields via TYPO3_CONF_VARS, e.g. `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT']['pagePath']['segTitleFieldList'] = 'tx_realurl_pathsegment,alias,title,uid';`

Any chance this option will be added to version 2.x?

ddiebel:
just a workaround:  you can override it in realurl_conf.php:

```
<?php
$GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] .= 'field1,field2';

\DmitryDulepov\Realurl\EncodeDecoderBase::$pageTitleFields = array('field1','field2','title','uid');
```

adrienjacob:
Thanks a lot for the workaround !

AlphawolfWMP:
Any update on this?

DenisMir:
Thanks for the workaround. An official update for this would be awesome.

PR #439, opened 2017-03-16
[FEATURE] Configurable page title field list like "segTitleFieldList"

birgerstoeckelmann:
Hi Dmitry,
this might be a solution to https://github.com/dmitryd/typo3-realurl/issues/199 . The hard coded property `pageTitleFields` is replaced by a configurable list via extension configuration.
If the list is not set, the dafault one will be used.
A new public static method gives access to this list and is also used in the DataHandler hook.
Maybe you find the time for having a look into it ;)


dmitryd:
This has to be reworked because the same thing will be unserialized too many times. Default values should be an array at the top of the class for easy adjustments (same static as it was before). The whole thing should be initialised in the constructor just once.

PR #446, opened 2017-03-22
[TASK] Trim multiple slashes for URLs with additional params

astehlik:
The trimming of multiple slashes at the end of a speaking URL
is now performed before adding remaining URL parameters.

This cleans up multiple slashes at the end of URLs that have
additional GET parameters:

/my/url/with/params///?param1=value1

is now cleaned up and looks like:

/my/url/with/params/?param1=value1


dmitryd:
Nice catch!

Issue #449, opened 2017-03-29
[BUGFIX] Unique aliases shouldn't be used when still active


frans-beech-it:
The check in `EncodeDecoderBase::createUniqueAlias()` only checks existing unique aliases with `expires=0` but it should check all active unique aliases as there can still be url's present in `tx_realurl_urldata` and even when not the unique alias could still be rebuild by the encoder/decoder.

PR #450, opened 2017-03-29
[BUGFIX] Don't re-use unique aliases that are still active

frans-beech-it:
Fixes: #449


frans-beech-it:
tnx for adding this :+1:

Issue #270, opened 2016-09-21
Core: Error handler (FE): PHP Warning: strpos(): Empty needle in typo3conf/ext/realurl/Classes/Decoder/UrlDecoder.php line 1057


abma:
i have redirect_goodUpperDir enabled and when accessing a bad url i get an endless redirect and in the syslog the above error is shown.

can be reproduced with:

/de/someurl//someur/

/de/someurl/ must be a valid url.

abma:
as workarround i added a redirect which removes double slashes from an url:

RewriteRule "^/(.+)\/\/(.+)$"  "$1"  [R=301]

dmitryd:
This would be nearly impossible to solve if there are more than one empty segment. Your best way would be to avoid such situations. Do your configuration in a way that does not create empty segments.

PR #451, opened 2017-03-30
[BUGFIX] 'redirect_goodUpperDir' with multiple slashes or bad parts in good URL (#437)

### LINKED ISSUE #437: PHP Warning: strpos(): Empty needle in realurl/Classes/Decoder/UrlDecoder.php line 1095
This is the same problem as #270 

realurl_conf.php
```
$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT'] = [
	'init' => [
		'appendMissingSlash' => 'ifNotFile',
		'postVarSet_failureMode' => 'redirect_goodUpperDir'
	],
	'preVars' => [
		[
			'GETvar' => 'L',
			'valueMap' => [
				'de' => '0',
				'en' => '1',
				'fr' => '2',
				'nl' => '3'
			],
		]
	],
	'fileName' => [
		'index' => [
			'index.html' => [
				'keyValues' => []
			]
		]
	]
];
```

If you now have a page `Impressum` which is available as `http://www.mydomain.org/de/impressum/index.html` and call

`http://www.mydomain.org/de/impressum//index.html` in your browser, you get a message about an endless redirect and an entry in `sys_log`: `Core: Error handler (FE): PHP Warning: strpos(): Empty needle in typo3conf/ext/realurl/Classes/Decoder/UrlDecoder.php line 1095` 

Each added `/` to the url adds one entry more on each request.

Using four slashes `http://www.mydomain.org/de/impressum////index.html` redirects to `http://www.mydomain.org/dimpressu///index.html` where some kind of mangling with the language parameter happens additionally.

This works up to eight slashes `http://www.mydomain.org/de/impressum////////index.html` where eight entries are written to `sys_log` for one request.

Using nine slashes `http://www.mydomain.org/de/impressum/////////index.html` results in a 404 page with the content `File not found.`

Using more than nine slashes results in a 500 error on my server.

While I understand it's not realurl's job to cope with empty parameter configurations added by extensions this issue needs to be addressed in my opinion. You can flood the `sys_log` table in the database of a system with that configuration. There must not be a endless redirect issued be realurl and requesting any url on a page with realurl must not create an PHP error.

sgrossberndt:
When using `redirect_goodUpperDir` handle multiple slashes in a requested URL by replacing all ocurrences of adjacent slashes by a single slash instead of logging a PHP warning and redirecting to the same wrong URL.

`good/good//index.html` redirects to `good/good/index.html`
`good/good//bad///index.html` redirects to `good/good/bad/index.html`, which redirects to `good/good/index.html`

Also search for the bad URL part starting from the back instead of from the front of the string. This lead to wrong URLs if the bad URL part was e.g. a single character which is also part of the valid base URL:

If e.g. the valid URL is 
`http://example.org/en/imprint/index.html`
and 
`http://example.org/en/imprint/in/index.html`
is requested, without the patch the bad part `in` is not replaced where it is actually wrong but in the word `imprint`, so you got redirected to 
`http://example.org/en/impt/in/index.html`.

By starting to search from the back the correct bad part is found and you get redirected to to correct URL `http://example.org/en/imprint/index.html`

Fixes #437, #270

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -1089,7 +1089,7 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 				} while ($replaced > 0);
 			} else {
 				// There is a unrecognized postVarSetKey
-				$badPathPartPos = strpos($this->originalPath, $badPathPart);
+				$badPathPartPos = strrpos($this->originalPath, $badPathPart);
 				if ($badPathPartPos > 0) {
 					// We also want to get rid of one slash
 					$badPathPartPos--;
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -1079,14 +1079,24 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 		if ($failureMode == 'redirect_goodUpperDir') {
 			$nonProcessedArray = array($postVarSetKey) + $pathSegments;
 			$badPathPart = implode('/', $nonProcessedArray);
-			$badPathPartPos = strpos($this->originalPath, $badPathPart);
 			$badPathPartLength = strlen($badPathPart);
-			if ($badPathPartPos > 0) {
-				// We also want to get rid of one slash
-				$badPathPartPos--;
-				$badPathPartLength++;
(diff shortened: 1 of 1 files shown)


dmitryd:
Thank you :)

PR #461, opened 2017-04-20
[BUGFIX] add base L config in realurl_autoconf.php even if no languages

### LINKED ISSUE #460: Realurl appends unwanted L and cHash url parameters
Realurl appends L and cHash parameters if a domain has no preVars configured for L and an L=0 query parameter is passed.

It happens in method addRemainingUrlParameters() and causes an url that should be like:
pagePath/value_alias/
to be
pagePath/value_alias/?cHash=3ba47534bc69d5fc7b7e2d86c7585bac&L=0

webian:
This partially fixes #460 in the case that auto configuration is used.

Maybe a base L config for preVars should be added always to the config array to have a common config base.

### DIFF
--- a/Classes/Configuration/AutomaticConfigurator.php
+++ b/Classes/Configuration/AutomaticConfigurator.php
@@ -80,6 +80,14 @@ class AutomaticConfigurator {
 	 * @return void
 	 */
 	protected function addLanguages(array &$configuration) {
+		$configuration['preVars'] = array(
+			0 => array(
+				'GETvar' => 'L',
+				'valueMap' => array(
+				),
+				'noMatch' => 'bypass'
+			),
+		);
 		if (version_compare(TYPO3_branch, '7.6', '>=')) {
 			$languages = $this->databaseConnection->exec_SELECTgetRows('t1.uid AS uid,t1.language_isocode AS lg_iso_2', 'sys_language t1', 't1.hidden=0 AND t1.language_isocode<>\'\'');
 		}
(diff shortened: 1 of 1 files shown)


liayn:
Disagree here. If it is a site with a single language, I don't want to see an L parameter anywhere.

dmitryd:
I do not see any reason to add unused variables to the configuration. This will not be added.

vmans:
In principle I agree to Dmitry. It may be a bug in the Core?

But for a clear operation: can't RealURL be improved with a checkbox in the extension configuration, that indicates if it's a single language site or not. Or for that matter, if L must be avoided or not. That prevents the consequence Dmitry addressed above, I guess. And it's an easy helpful solution for any level of experience.

dmitryd:
RealURl already detectes languages and adds them if there are more than one.

vmans:
I get that. But if there is only one language, the L=0 is kept alive in the URI. That's what we don't need. Not saying that is caused by RealURL. Saying that it can be solved easily there. Perhaps I'm wrong, but it's a suggestion that would help me and many others I guess.

dmitryd:
> Not saying that is caused by RealURL. Saying that it can be solved easily there

No. It should be solved by fixing whatever caused it. RealURL will not fix user errors.

jansalleine:
Just for completeness sake, if anyone comes from google: 
I dug a little deeper into this issue and it may be most likely caused by the Typo3 indexed_search extension which adds the 'L' parameter to all URLs in SearchController.php, protected function linkPage($pageUid, $linkText, $row = [], $markUpSwParams = []). I checked a couple Typo3 installations from my job and all that had indexed_search running had this issue and those who had another search or no search at all didn't have this issue.

PR #464, opened 2017-04-25
[TASK] Make exception text more descriptive

liayn:
If multiple root pages for a domain are found an exception is thrown.
Enhance the text to inform the user that the issue was multiple root
pages and not anything else.


dmitryd:
Thank you!

PR #467, opened 2017-04-25
Add Hook to urlsegment and pageidentifier comparison

stiXits:
Inserted a hook when checking if an url segment matches a page identifier (title, path segment etc.).
Effectively new disjunctive conditions can be added to the decision whether an url segment matches a page identifier.
Cobined with a different extension I use this hook to restore the realurl 1.x functionality of accepting "-" as well as "_" as seperation characters. But this hook may also be interesting for other functionality.

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -340,8 +340,17 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 			if ($this->detectedLanguageId > 0 && !isset($page['_PAGES_OVERLAY']) && (empty($languageExceptionUids) || !GeneralUtility::inList($languageExceptionUids, $this->detectedLanguageId))) {
 				$page = $this->pageRepository->getPageOverlay($page, (int)$this->detectedLanguageId);
 			}
+
+                        // initiate seperation character hooks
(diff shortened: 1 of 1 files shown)


dmitryd:
This will need code reformatting at least...

Classes/Decoder/UrlDecoder.php
```diff
@@ -340,8 +340,17 @@ protected function createPathCacheEntry($segment, array $pages, array &$shortcut
 			if ($this->detectedLanguageId > 0 && !isset($page['_PAGES_OVERLAY']) && (empty($languageExceptionUids) || !GeneralUtility::inList($languageExceptionUids, $this->detectedLanguageId))) {
 				$page = $this->pageRepository->getPageOverlay($page, (int)$this->detectedLanguageId);
 			}
+
+                        // initiate seperation character hooks
+                        $conditionHooks = array();
+                        if (is_array($GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['linkvalidator']['checkLinks'])) {
```
dmitryd: linkValidator???

dmitryd:
Rejected, because looks untested: checks for linkhandler hooks but loops over 'spaceCharacter', which is a bad name for the hook.

PR #473, opened 2017-05-04
Possibly remove leading slash from siteScript

RaphaelBergmann:
There might be a leading slash in $this->siteScript, so isSpeakingUrl() would return true for e.g. /index.php?id=54&L=0 and then show the page not found-error 'Reason: Segment "index.php" was not a keyword for a postVarSet'.


dmitryd:
I never saw a case when there can be a slash before. Please, describe how I can reproduce the issue.

RaphaelBergmann:
Oh, I'm sorry, this one's on me. I set [SYS][reverseProxyPrefixSSL] = /news/ instead of [SYS][reverseProxyPrefixSSL] = /news and since everything else was working properly I did not figure it out immediately...

PR #477, opened 2017-05-09
Sys language uid hook

stiXits:
Added a Hook that runs before the decision whether a page language overlay is fetched before evaluating page identifiers and url segments.

This can, for example, be used to evaluate the sys_language_uid setting to determine a page's currently targeted language. The sys_language_uid setting is often used to change the default language within the page tree.

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -313,6 +313,38 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 		return $result;
 	}
 
+        public static function getSysLanguageUid( &$page )
+        {
+		//\TYPO3\CMS\Extbase\Utility\DebuggerUtility::var_dump("chacking page id:" . $page['uid'] . " title: " . $page['title']);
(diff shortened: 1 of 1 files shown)


Classes/Decoder/UrlDecoder.php
```diff
@@ -313,6 +313,38 @@ protected function convertAliasToId(array $configuration, $value) {
...
+                //\TYPO3\CMS\Core\Utility\GeneralUtility::devLog("Retrieving sys_language_uid: ", "realurl", 1, array(0));
+                $type = 0;
+                $name = 'TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController';
+                /*if( !is_object( $GLOBALS['TT'] ) )
+                {
+                    $GLOBALS['TT'] = new \TYPO3\CMS\Core\TimeTracker\TimeTracker();
+                    $GLOBALS['TT']->start();
+                }*/
+
+                $GLOBALS['TSFE'] = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance( $name,  $GLOBALS['TYPO3_CONF_VARS'], $page['uid'], $type );
```
dmitryd: None of such code will come to realurl because it is a hack.

PR #478, opened 2017-05-10
[Bugfix] Proper crossdomain linking detection

adrienjacob:
After testing, it seems isLinkingAcrossDomains (introduced in https://github.com/dmitryd/typo3-realurl/commit/1dbbe84da5d5bd419f2418c2f73fcb6c853b395c) always returns true... This is a suggested fix, as well as a small change on the similar function isPageInRootlineOfRootPage in UrlDecoder so that both functions are consistent with each other.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -1128,7 +1128,7 @@ class UrlEncoder extends EncodeDecoderBase {
 	protected function isLinkingAcrossDomains() {
 		$result = true;
 
-		foreach (array_reverse($this->tsfe->rootLine) as $page) {
+		foreach ($this->tsfe->rootLine as $page) {
 			if ($page['uid'] == $this->rootPageId) {
 				$result = false;
 				break;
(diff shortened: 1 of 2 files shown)


Classes/Encoder/UrlEncoder.php
```diff
@@ -1128,7 +1128,7 @@ protected function initializeConfiguration() {
 	protected function isLinkingAcrossDomains() {
 		$result = true;
 
-		foreach (array_reverse($this->tsfe->rootLine) as $page) {
+		foreach ($this->tsfe->rootLine as $page) {
```
dmitryd: How will removal of `array_reverse` help? Please, explain.

dmitryd: Explanation for the current code.

`$this->tsfe->rootLine` is reversed. Without `array_reverse`, `foreach` will start from the currently generated page and move up to the tree. With `array_reverse` we start from the root pages (globe level).

So I am trying to understand what will this change actually change, except walking more pages.

adrienjacob:
Hmmm... very strange. After your comment I realised my suggested change wasn't supposed to have any effect indeed. And in fact, after testing again, it seems everything looks good (config.absRefPrefix is taken into account). Don't know what happened in the meantime... Anyway, case closed !

Classes/Encoder/UrlEncoder.php
davidjakob: Removing `array_reverse` would allow crossdomain detection to work with nested domains.

PR #490, opened 2017-06-06
A primary key for the table tx_realurl_uniqalias_cache_map

ryno83:
When using Percona XtraDB Cluster in strict mode = ENFORCING or MASTER (highly recommanded) it's not permit to have a table without a primary key. Then I'm proposing to add a primary key for the table tx_realurl_uniqalias_cache_map.


dmitryd:
A simple integer autoincrement field will be better.

dmitryd:
May be, I should tell why.

Firsts, I do not like to duplicate indexes. Secondly, `uid` is a de facto standard for the PK in TYPO3.

PR #491, opened 2017-06-07
[BUGFIX] Fix ignored `utm_*` get-parameters list

romm:
By default, TYPO3 stores the following values in the configuration `cHashExcludedParameters`: `utm_source, utm_medium, utm_campaign, utm_term, utm_content`.

Currently, RealURL accepts any `utm_[a-z]` parameter.

This commit sets this list in the RealURL ignored get-parameters list, to prevent any `utm_*` parameter that is not supported from being added.

### DIFF
--- a/Classes/Configuration/ConfigurationReader.php
+++ b/Classes/Configuration/ConfigurationReader.php
@@ -85,7 +85,7 @@ class ConfigurationReader {
 	 */
 	protected $defaultValues = array(
 		'cache/banUrlsRegExp' => '/tx_powermail_pi1%5Baction%5D=create|tx_solr|tx_indexedsearch|tx_kesearch|(?:^|\?|&)q=/',
-		'cache/ignoredGetParametersRegExp' => '/^(?:gclid|utm_[a-z]+|pk_campaign|pk_kwd|TSFE_ADMIN_PANEL.*)$/',
+		'cache/ignoredGetParametersRegExp' => '/^(?:gclid|utm_(source|medium|campaign|term|content)|pk_campaign|pk_kwd|TSFE_ADMIN_PANEL.*)$/',
 		'fileName/acceptHTMLsuffix' => TRUE,
 		'fileName/defaultToHTMLsuffixOnPrev' => FALSE,
 		'init/appendMissingSlash' => 'ifNotFile,redirect[301]',


fedir:
I confirm, this bug is real and needs to be merged ASAP.

filoucrackeur:
I confirm too

dmitryd:
@fedit This is not related to security at all.

fedir:
@dmitryd I made Denial-of-service attack on one of our private sites using this vulnerability. Basically : it replaces cached version of the page by an error page. So all attacked pages are not available anymore.

dmitryd:
@fedir Do you expect normal users to go and append `utm_crazyparameters` to your urls and enjoy seeing 404s? Really?

dmitryd:
I have even better option for DoS: add a couple of random parameters to the URL, make sure that `pageNotFoundOnCHashError` is `true` and add `cHash=111` to the URL. There will be a 404. DoS?

In other words, synthetic situations (those that do not happen in a real life) are not very interesting to me. If somebody wants to craft a URL and see a 404, I do not see a problem. He wanted, he made it. Cool. My problem is only when regular people see 404s.

romm:
Hi @dmitryd, thank you for taking the time to analyze the issue and merge the PR!

Concerning you last comment, I would add that putting a random parameter like `utm_crazyparameters` **does not crash the actual page** being requested, but the default one without particular parameter, that's where the issue is (technically, RealURL fetches the wrong cache entry because cache has been corrupted, leading to a wrong cHash and making TYPO3 "displeased").

So your last example does not apply to this particular situation: one could cause a page crash for **other users**!

Anyway, this PR should solve this problem and I'm glad you accepted it. Thanks a lot for all the work you did. 👍

fedir:
@dmitryd Yes, as @romm said, the original page is not available anymore after the attack, basically, any page of Your site could be transformed into 404 for all users.

dmitryd:
@romm 
> RealURL fetches the wrong cache entry because cache has been corrupted, leading to a wrong cHash and making TYPO3 "displeased"

I do not think it works like this. Suppose you have a URL like `/path/to/page/news/detai/my-news-title/`. The original URL will look something like `/index.php?id=123&tx_news[showUid=456&cHash=78901234567890`. The combinations of those two goes to the realurl cache. Now if you add a `utm_blahblah` parameter, the following happens:

* the url becomes `/path/to/page/news/detai/my-news-title/?utm_blahblah=test`
* realurl will strip the parameter for the cache lookup
* realurl will find `/path/to/page/news/detai/my-news-title/` in the cache
* realurl will produce the url, which looks like `/index.php?id=123&tx_news[showUid=456&cHash=78901234567890&utm_blahblah=test`
* TYPO3 will calculate cHash including `utm_blahblah`, see that cHash does not match
* if you have `pageNotFoundOnCHashError` set to true (as you should!), you will have a 404 and no problems with page cache. See [TypoScriptFrontendController::makeCacheHash()](https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L2133)
* if you have `pageNotFoundOnCHashError` set to false, the user will see the page but the page will not be cached. See same method as above.

In both cases the cache will not be affected. So the scenario you describe may not happen in a clean TYPO3 installation.

@fedir Cannot confirm this with a clean TYPO3. See my analysis above. I believe your system somehow modifies TYPO3 and makes the problem happen.

Elypson:
I could reproduce taking down a page as well.

For example, if you have a language switch that creates a URL based on the called URL AND your page cache as well as the realurl path cache are empty, then the new language switch URL will be created based on the cHash value that realurl uses. However, if the URL is later decoded by any user, realurl will find the cached path with the cHash value, take it to TYPO3 but TYPO3 will not find it because it has ignored other parameters than realurl.

Do you think, we should have a different configuration where this error is prevented?

PR #492, opened 2017-06-09
Fix redirect_goodUpperDir bugs

sgrossberndt:
http://example.org/bad1/bad2/bad3/ and http://example.org/good/bad1/bad2/bad3/ creates to a wrong redirect.

If `$pathSegments` contains more than one element the operation `$nonProcessedArray = array($postVarSetKey) + $pathSegments` leads to only the last element being appended to the `$badPathPart`, e.g. http://example.org/bad1/bad2/bad3/ leads to `$badPathPart = 'bad1/bad3';` which will not be found in `$this->originalPath`. `$nonProcessedArray = array_merge(array($postVarSetKey), $pathSegments)` correctly identifies `$badPathPart ='bad1/bad2/bad3';` creating the correct redirect.

Currently the badPathPart handling for two or more adjacent slashes is triggered even if there is only one slash, using `strpos($badPathPart, '//')` yields the correct behaviour.


dmitryd:
Thanks!

PR #508, opened 2017-07-21
[BUGFIX] Removed wrong typecast

maximilian-walter:
Don't cast the field-name to an integer. Otherwise path-segments of translated records (non-pages) will not be automatically updated if the record is changed.


dmitryd:
Thank you!

PR #517, opened 2017-08-09
[BUGFIX] Drop deprecated GeneralUtility::requireOnce call

lochmueller:
Drop the call of GeneralUtility::requireOnce and replace it with the PHP native require_once

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -26,7 +26,7 @@ if (!function_exists('includeRealurlConfiguration')) {
 
 		$realurlConfigurationFile = trim($configuration['configFile']);
 		if ($realurlConfigurationFile && @file_exists(PATH_site . $realurlConfigurationFile)) {
-			\TYPO3\CMS\Core\Utility\GeneralUtility::requireOnce(PATH_site . $realurlConfigurationFile);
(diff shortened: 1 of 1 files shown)


dmitryd:
Not enough. If you see the original function, it makes certain globals available:

```php
    public static function requireOnce($requireFile)
    {
        self::logDeprecatedFunction();
        // Needed for require_once
        global $T3_SERVICES, $T3_VAR, $TYPO3_CONF_VARS;
        require_once $requireFile;
    }
```

Your solution changed the behavior and adds a regression.

lochmueller:
Obsolete.

DanielRuf:
> Your solution changed the behavior and adds a regression.

https://github.com/dmitryd/typo3-realurl/pull/607

DanielRuf:
https://github.com/dmitryd/typo3-realurl/commit/ab6a5cf5c5fed1bf804ed1119a469ed72406c0c6

DanielRuf:
Also 9.0.0 is not supported so it is safe to revert this from `require_once` to `\TYPO3\CMS\Core\Utility\GeneralUtility::requireOnce`.

PR #525, opened 2017-08-24
Remove pointless explicit transaction

RasmusWernerLarsen:
In the limitTableRecords function a periodic cleanup is triggered, that
comes down to a select and delete function, which is run inside an
explicit transaction. Given that we only have a single data modifying
operation that doesn't seem to serve any purpose.

The reason for our interest in this change, is that we're seeing some heavy periodic locking activity hitting the tx_realurl_urldata table, and the cause seems to be this transaction.
It's important to remember that, when using MySQL and starting a transaction, you're by default setting the database to use the "Repeatable Read" isolation mode. Using "repeatable read" means that any select you perform will set locks for the rows you access, until the transaction is completed.

I'll note that in other parts of the code you're using transactions in places where it makes sense, but in those cases you should seriously consider lowering the isolation level to READ-COMMITTED, to prevent excessive locking. I don't see any parts of RealURL that would need "Repeatable Read".

### DIFF
--- a/Classes/Cache/DatabaseCache.php
+++ b/Classes/Cache/DatabaseCache.php
@@ -424,7 +424,6 @@ class DatabaseCache implements CacheInterface, SingletonInterface {
 	protected function limitTableRecords($tableName) {
 		$cleanedUp = false;
 		if ((mt_rand(0, mt_getrandmax()) % 5000) == 0) {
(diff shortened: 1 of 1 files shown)


dmitryd:
Thanks!

PR #528, opened 2017-09-08
cache not updated if a page is renamed through a workspace

bill622:
I know workspaces are less used, but where I work we really need them... So this code solve a cache bug when a page is being renamed in a workspace then published to live. The Url cache is now updated if a version change occurs.

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -1122,8 +1122,9 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 	protected function handleNonExistingPostVarSet($pageId, $postVarSetKey, array &$pathSegments) {
 		$failureMode = $this->configuration->get('init/postVarSet_failureMode');
 		if ($failureMode == 'redirect_goodUpperDir') {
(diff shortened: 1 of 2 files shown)


dmitryd:
RealURL does not work with workspaces. This is officially documented.

Classes/Decoder/UrlDecoder.php
```diff
@@ -1122,8 +1122,9 @@ protected function handleFileNameMappingToGetVar(&$fileNameSegment, array &$getV
 	protected function handleNonExistingPostVarSet($pageId, $postVarSetKey, array &$pathSegments) {
 		$failureMode = $this->configuration->get('init/postVarSet_failureMode');
 		if ($failureMode == 'redirect_goodUpperDir') {
-			$nonProcessedArray = array($postVarSetKey) + $pathSegments;
+			$nonProcessedArray = array_merge( array($postVarSetKey), $pathSegments );
 			$badPathPart = implode('/', $nonProcessedArray);
+			$badPathPartPos = strrpos($this->originalPath, $badPathPart);
```
dmitryd: Is this needed here? It is not used outside of `else` block below.

PR #532, opened 2017-09-18
[BUGFIX] Fix #531: Wrong behaviour when changing cHash relevant settings

### LINKED ISSUE #531: Wrong behaviour when changing cHash relevant settings
It happened that I changed `$TYPO3_CONF_VARS['FE']['cHashIncludePageId']` on a website. As this will create a new cHash value, RealUrl now finds two rows of the same URL in `tx_realurl_urldata`.

While the newer result would be the correct one, RealUrl picks the first record which is (because of the default sorting) the older record (which has the lower uid).

I suggest to fix this by sorting by uid in descending order. This will most likely cost some additional milliseconds, but it fixes potential issues with cHash which are most of the time hard to find.

The alternative solution would be to make sure that no duplicate records of the same URL can be stored in `tx_realurl_urldata`...

stucki:
see #531


dmitryd:
Closing because the cache has to be cleared manually in such cases according to TYPO3 docs.

PR #534, opened 2017-09-22
Update UrlDecoder.php

TokeHerkild:
When DB::exec_SELECTgetRows() returns non traversable method throws an InvalidOffset Exception and the realurl cache for requested page is invalidated in such a way that the page cannot be accessed until a flush::all command is issued.


dmitryd:
This means your database is not up to date. There is no need for this check. Update the database.

PR #544, opened 2017-10-09
[Feature] Append (Copy) to language overlay titles

adrienjacob:
This PR appends "(copy x)" not only to the title of the copied pages, but also to the title of the copied language overlays. This solves an issue we often have, and documented here : https://forge.typo3.org/issues/76038
When copying a given page with several language overlays, the copied page can have "(copy 1)" appended, but not language overlays, resulting in many URL conflicts - not always properly identified by editors since language overlays don't appear in the menu tree.

As this core "bug" is mostly an issue when using realurl, I assumed it would make sense to apply the fix within realurl itself. Besides I expect it will have a rather low priority for the core team.

It relies on two hooks : `processCmdmap_preProcess` before the copy is performed (to determine the need for a title change), and `processCmdmap_afterFinish` to apply the title change (once we know the Id of the newly created page). It is adapted from the core code managing the "(copy)" tags within DataHandler.

### DIFF
--- a/Classes/Hooks/DataHandler.php
+++ b/Classes/Hooks/DataHandler.php
@@ -43,6 +43,16 @@ class DataHandler implements SingletonInterface {
 
 	/** @var \TYPO3\CMS\Dbal\Database\DatabaseConnection */
 	protected $databaseConnection;
+	
+	/**
+	 * For the list of modified language overlay titles
+	 *
+	 * @var array
+	 */
+	private $newTitles = [];
+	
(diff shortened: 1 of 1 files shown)


dmitryd:
I am sorry but I do not think I will integrate that. There are several reasons for this:

1. Since this is a core bug, the proper place is to fix it in the core. Anybody can make a fix for the core.
1. Realurl does not modify page titles because it is user data. It only converts urls from the user data. This code would change the behavior and can cause unwanted and unexpected results for users.

adrienjacob:
Okay, makes sense...

Issue #320, opened 2016-11-03
Shortcut handling seems to be missing from version 2


tarleb:
Version 1 of the extension allowed to control shortcut handling via the `dontResolveShortcuts` config option.  It involved the functions `resolveShortcut` and `resolveShortcuts` in `class.tx_realurl_advanced`, making it possible to substitute the URL of a shortcut page with the URL of the shortcut's target page.  It appears that this is no longer possible.

To give an example:

```
|-- Some Content
| \
|  ` Important Page
|
|-- Footer menu
 \
  ⋮ other menu pages
  ` Shortcut to Important Page
```

Realurl version 1 would generate a link to *Important Page* in the footer menu, with realurl version 2.1.4 the link points to the shortcut page instead (which then does a redirect to *Important Page*, causing an extra round-trip and with some configs even a redirect loop with itself).

That's a pitty, as that feature is very useful for SEO.  This is also not mentioned in the documentation, neither the "differences between v1 and v2" nor the "upgrading to v2" wiki pages make any mention of this.   This is a breaking change causing unexpected behavior and I'd hence qualify it as a bug.

Are there workarounds for this?  If not, would a pull-request which re-implements this be welcome?

dmitryd:
realurl v2 is not 100% compatible with v1. You can try making a PR but you have to make sure it works across domains, with all kind of page types, mount points, etc.

tarleb:
Thank you for the feedback.  Would you kindly add this difference to the wiki? It might safe people from running into that issue.

We'll take the old `resolveShortcut` code as the basis and might be able to compile a PR in the not-to-distant future.

tarleb:
I'm closing this for now, as I'm not sure if I'll be getting the resources to code a general fix.

PR #550, opened 2017-10-12
[FEATURE] Resolve shortcut destination

the-coding-owl:
According to issue #320, I would like to reintegrate shortcut resolving into the realurl feature pool.
Therefore the pagePath/dontResolveShortcuts configuration has been added again and set to TRUE by default, to not break existing realurl 2.x installations.


EdRoxter:
On a quick glimpse, this looks like a good solution and I'd definitely second the request to take this back into RealURL

Issue #549, opened 2017-10-12
Bad L parameter


Guigue79:
Hi
When users or bots try to set Language variable to fr
I have this error
Thu, 12 Oct 2017 09:46:25 +0200 [ERROR] request="eb2a025b3e3ec" component="DmitryDulepov.Realurl.Encoder.UrlEncoder": Bad "L" parameter ("fr") was detected by realurl......
Logs are very very very large and saturate my disk
I try to ignore this param with config.linkVars = L(0-1) but it does not work

osahner:
Same here. Logging seems to be messed up this release. One Line has more than 3mb!

responseinformationsdesign:
RealURL also seems to ignore [SYS][systemLogLevel].

Errors get written in TYPO3-Log even [SYS][systemLogLevel] was set to only log fatal errors.

hannesbochmann:
@responseinformationsdesign 
realurl uses the logging system ($TYPO3_CONF_VARS['LOG']) of TYPO3 not the simple system log. System log level has no effect on the logging system. You would need to remove the default FileWriter for log level WARNING after you copied it to the log configuration for log level ERROR. This way you should only have errors in typo3temp/logs/typo3.log. By the way this is no recommended. Warnings shouldn't be ignored. ;-)

bavarianbytes:
Same here

fazzyx:
We have the same situation after upgrading realurl to 2.3. 
TYPO3 7.6

Batman777:
Typo3 - 6.31
Same problem with huge log files. Will there be a fix in the near future?

> You would need to remove the default FileWriter for log level WARNING after you copied it to the log configuration for log level ERROR.

I did not get that. Could you please try in other words?

hannesbochmann:
@Batman777 
Put that in your AdditionalConfiguration.php:
```
$GLOBALS['TYPO3_CONF_VARS']['LOG']['writerConfiguration'][\TYPO3\CMS\Core\Log\LogLevel::ERROR]['TYPO3\\CMS\\Core\\Log\\Writer\\FileWriter'] = $GLOBALS['TYPO3_CONF_VARS']['LOG']['writerConfiguration'][\TYPO3\CMS\Core\Log\LogLevel::WARNING]['TYPO3\\CMS\\Core\\Log\\Writer\\FileWriter'];
[...]

PR #551, opened 2017-10-24
Remove logging of debug information when bad "L" param was discovered

dwinkelbauer:
As some search engine bots visit our site with an invalid "L" param every few hours, our typo3temp logfile grew up to several GB. This is caused by logging the debug_backtrace() call, which should be removed.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -1586,7 +1586,7 @@ class UrlEncoder extends EncodeDecoderBase {
 				addslashes($sysLanguageUid)
 			);
 			$this->tsfe->set_no_cache($message);
-			$this->logger->error($message, debug_backtrace());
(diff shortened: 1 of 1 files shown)


tantegerda1:
Related issue: #549

LeoniePhiline:
+1. debug_backtrace() is not helping anyone and generates megabytes of data for each log entry.

DrillSergeant:
That would be a good thing to merge this. +1

dmitryd:
@LeoniePhiline 
> +1. debug_backtrace() is not helping anyone and generates megabytes of data for each log entry.

Don't talk for anyone, please. I had two clients, whom it actually helped.

dmitryd:
Fixed [here](https://github.com/dmitryd/typo3-realurl/commit/4d1c5f5b18bc3ea485723a9801a83f3a8c8cc1ff)

Issue #541, opened 2017-10-05
RealURL adds a multiple/wrong entry to tx_realurl_urldata for foreign root pages on multidomain websites


bihor:
I have the problem that I can visit the homepage of a foreign language only once.
When I visit it again, I get a 404-error-page.
I guess the problem is that RealURL creates multiple entries for one page in the tx_realurl_urldata table and only the first one is correct.
I don´t know why RealURL does not use the correct first entry.
My realurl_conf.php looks like this:

![realurl_lang_problem](https://user-images.githubusercontent.com/8191408/31233589-4bbbe59c-a9ed-11e7-8f9d-c6a83490d7d5.jpg)
![realurl_lang_problem2](https://user-images.githubusercontent.com/8191408/31234168-7b83cdf2-a9ee-11e7-93ab-78270608ba1c.jpg)

`$TYPO3_CONF_VARS['EXTCONF']['realurl']['_DEFAULT'] = array(
	...
	'preVars' => array(
		array(
			'GETvar' => 'L',
			'valueMap' => array(
				'fr' => '3',
				'pt' => '2',
				'es' => '1',
			),
			'noMatch' => 'bypass',
		),
		array(
			'GETvar' => 'no_cache',
			'valueMap' => array(
				'nc' => 1,
			),
			'noMatch' => 'bypass',
		),
	),
);

$TYPO3_CONF_VARS['EXTCONF']['realurl']['www.my2domain.org'] = $TYPO3_CONF_VARS['EXTCONF']['realurl']['_DEFAULT'];
$TYPO3_CONF_VARS['EXTCONF']['realurl']['www.my2domain.org']['pagePath']['rootpage_id'] = 513;
$TYPO3_CONF_VARS['EXTCONF']['realurl']['www.my2domain.org']['errorPage'] = 600;`

I have the problem on 2 different servers...
I both images is only the first entry correct! The second entry use the rootpage_id of another domain, thats why a get a 404-error-page.

bihor:
PS: used RealURL version: 2.2.1

dmitryd:
It seems that you did not set up languages as described in the realurl documentation.

Please, [ask for further help in Slack](https://github.com/dmitryd/typo3-realurl/wiki/Getting-help). Support is not offered in the bug tracker.

PR #554, opened 2017-11-02
Update UrlEncoder.php

cristianbuja:
FIX issue #541 (https://github.com/dmitryd/typo3-realurl/issues/541)


dmitryd:
This pull request does not fix anything but instead mitigates user's misconfiguration by making realurl avoid user errors. This will not be integrated. User is responsible for proper configuration of his system.

Issue #504, opened 2017-07-11
[BUG] Exception thrown if $GLOBALS['TYPO3_CONF_VARS']['FE']['cacheHash']['excludedParameters'] is not initialized


georg-tiefenbrunn:
Since 8cfb6f02 EXT:realurl causes an exception when installing TYPO3 CMS/flushing caches using TYPO3 Console.

> [Sub-process exception: TYPO3\CMS\Core\Error\Exception ]
 #1: Warning: in_array() expects parameter 2 to be array, null given in /var/www/html/typo3conf/ext/realurl/ext_localconf.php line 74
 thrown in file Classes/Error/ErrorHandler.php
 in line 68

Looks as if `$GLOBALS['TYPO3_CONF_VARS']['FE']['cacheHash']['excludedParameters']` is not initialized at that moment.

dmitryd:
The fix is not exactly right. If the array does not exist, it should not be used, that's all.

markuspoerschke:
@dmitryd I added this change here: #559 

If change in #559 is fine, we can close this Merge Request.

tbal:
A possible workaround till this bug is fixed in realurl is to initialize the array in LocalConfiguration.php :
```
    ...
    'FE' => [
        ...
        'cacheHash' => [
            'excludedParameters' => []
        ],
        ...
    ],
    ...
```

dmitryd:
This will never be "fixed" in RealURL because it is not a bug. Any configuration should be done only after TYPO3 is initialized. And TYPO3 has this option set already, so RealURL will not set it again.

See: https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Configuration/DefaultConfiguration.php#L1034

markuspoerschke:
@dmitryd For TYPO3 Version 7.6 this value is not present yet. Therefore this fix is only needed if TYPO3 <8 is used. Can you check #559?

PR #556, opened 2017-11-04
Fix PHP Notice: Undefined index: pageOverlayFields

tmotyl:
In PHP 7.0

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -46,8 +46,8 @@ $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['t3lib/class.t3lib_tcemain.php']['proc
 
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['realurl']['cacheImplementation'] = 'DmitryDulepov\\Realurl\\Cache\\DatabaseCache';
 
-$GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] .= ($GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] ? ',' : '') . 'tx_realurl_pathsegment,tx_realurl_exclude,tx_realurl_pathoverride';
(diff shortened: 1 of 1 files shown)


dmitryd:
I believe this is not necessary because it is a local issue. Normally realurl configuration is included after the `LocalConfiguration.php` and default system configuration is included. So this really should not be necessary.

liayn:
This is yet another issue which somehow is related to "missing configuration" 
See also #572 #504 #559

liayn:
More info here `pageOverlayFields` have been removed in TYPO3 v8

ext_localconf.php
```diff
@@ -46,25 +46,34 @@ function includeRealurlConfiguration() {
...
 
-$GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] .= ($GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] ? ',' : '') . 'tx_realurl_pathsegment,tx_realurl_exclude,tx_realurl_pathoverride';
-$GLOBALS['TYPO3_CONF_VARS']['FE']['pageOverlayFields'] .= ($GLOBALS['TYPO3_CONF_VARS']['FE']['pageOverlayFields'] ? ',' : '' ) . 'tx_realurl_pathsegment';
+if (empty($GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'])) {
+    $GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] = 'tx_realurl_pathsegment,tx_realurl_exclude,tx_realurl_pathoverride';
+} else {
+    $GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'] .= ',tx_realurl_pathsegment,tx_realurl_exclude,tx_realurl_pathoverride';
+}
+
+if (empty($GLOBALS['TYPO3_CONF_VARS']['FE']['pageOverlayFields'])) {
```
liayn: We should rather add a version check here, because this is v7 only.

PR #560, opened 2017-11-07
[FIX] reset encodedUrl for use with ext:crawler

samowitsch:
Hi @dmitryd. I have a fix in combination with this pull request
https://github.com/AOEpeople/crawler/pull/215. It is a simple fix for crawler usage with activated realurl option. Maybe you have another solution to reset the property encodedUrl.

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -126,6 +126,7 @@ class UrlEncoder extends EncodeDecoderBase {
 	 * @return void
 	 */
 	public function encodeUrl(array &$encoderParameters) {
+	    	$this->encodedUrl = '';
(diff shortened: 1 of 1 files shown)


samowitsch:
@dmitryd have you got time to look on it?

dmitryd:
Hm. Why is this needed? Encoder is not singleton. It is expected to be created each time when the new url is to be encoded. It will not work properly if it is reused. Crawler should create a new object each time when it tries to encode the URL.

samowitsch:
Hi. thx for feedback. I will discuss in https://github.com/AOEpeople/crawler/pull/215 and close this one later on.

samowitsch:
I close this PR. Problem could be solved in AOEpeople/crawler#215.

PR #559, opened 2017-11-06
Exception thrown if $GLOBALS['TYPO3_CONF_VARS']['FE']['cacheHash']['excludedParameters'] is not initialized

markuspoerschke:
This will fix #504

PR #561, opened 2017-11-07
[BUGFIX] fix missing cHash configuration issue

maritwho:
If the TYPO3 bootstrap mechanism wasn't executet completely yet,
the check for gclid in cacheHash->excludedParameters ends
up in an php error, because the array isn't generated yet.
This happens for example with helhum/typo3_console help
command.

To solve this, add a condition to check if the array even exists.


markuspoerschke:
Same as #559?

maritwho:
Ok, thank you for the hint, so I will close the pull request.

Issue #548, opened 2017-10-12
absRefPrefix on shortcut page


Guigue79:
Hi.
I have this error when users try to access to a shortcut page.
Thu, 12 Oct 2017 09:10:27 +0200 [WARNING] request="54759cfb33cc4" component="DmitryDulepov.Realurl.Encoder.UrlEncoder": config.absRefPrefix is not set! Please, check your TypoScript configuration! 
The config.absRefPrefix is set in the root page.

Patsone:
Same here. My TS contains config.absRefPrefix = / but I receive the same warnings filling up my logfile.

hannesbochmann:
The same here.

h3nn3s:
Jeb, same here. Looks like it is only fired if rootpage is a shortcut

bigahuna:
Hi, we can confirm that we also have this error if rootpage is a shortcut to another page.

DiablosOffens:
I've got the same problem and in combination with some SQL injection attacks (where the "L" parameter cannot be parsed by validateLanguageParameter in UrlEncoder.php) the logfile used up the whole disk space!

AndreasA:
I can also confirm this warning.
The issue is that checkPageForShortcutRedirect is executed prior to setAbsRefPrefix 
which means that the URL encoder is executed before TSFE->absRefPrefix is set which is OK here because TYPO3 takes care to append the current host information afterwards.

adrienjacob:
Hmmmm.... I also have this issue (but not the one with L parameter, maybe an issue with addQueryString or config.linkVars).
Anyway, this absRefPrefix warning seems impossible to avoid when using shortcuts : we cannot access protected property `TypoScriptFrontendController::$originalShortcutPage` to know if we are currently following a shortcut. So maybe we should just remove this warning logging ?

fazzyx:
@markuspoerschke There is a devlog switch, maybe depend the logging on it.

AndreasA:
@markuspoerschke I checked the code and in 7.6, 8.x and current master immediately after absRefPrefix is set the cObj in TSFE is created by executing newCObj. So you could also check regarding that.

ervaude:
You can deactivate the logging of warnings from this class with

```
$GLOBALS['TYPO3_CONF_VARS']['LOG']['DmitryDulepov']['Realurl']['Encoder']['writerConfiguration'] = [
        \TYPO3\CMS\Core\Log\LogLevel::WARNING => []
];
```

[...]

PR #563, opened 2017-11-09
Remove warning about missing absRefPrefix configuration

markuspoerschke:
The warning will be logged whenever an URL is created because of a bug within TYPO3.

See:
- #548 (https://github.com/dmitryd/typo3-realurl/issues/548)
- https://forge.typo3.org/issues/82939


dmitryd:
I do not think it is a good idea to remove the warning. It is there for a reason. I'd rather prefer a core fix for this.

dmitryd:
See https://github.com/dmitryd/typo3-realurl/commit/54c5233041f7ba272ac61a6c2c2f08950294c0c8

markuspoerschke:
Makes sense! Thanks for the feedback!

PR #565, opened 2017-11-16
[BUGFIX] #504: Prevent warning when $GLOBALS['TYPO3_CONF_VARS']['FE']…

MarkTro:
…['cacheHash']['excludedParameters'] is not set


dmitryd:
How do I reproduce the issue? In my default setup it is always set.

liayn:
Sounds like this could be related to #572

Issue #568, opened 2017-11-19
realurl_autoconf.php not evaluated when first time re-created


webian:
If I delete realurl_autoconf.php, the first time that it's re-created, its array is not evaluated so it's not put into `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']` that is still null after `require_once(PATH_site . TX_REALURL_AUTOCONF_FILE);` (https://github.com/dmitryd/typo3-realurl/blob/development/Classes/Configuration/ConfigurationReader.php#L299).

I guess that it's a PHP bug because the file exists at that moment, it's just that it is not evaluated by the `require_once` command.

dmitryd:
In your PR you just try to load the same file second time. This looks like a hack, not like a solution... There should be a proper fix for this.

rrrapha:
I can't explain.. but using require() instead of require_once() seem to do the trick.
@webian Can you confirm?

PR #569, opened 2017-11-19
[BUGFIX] Makes sure that realurl_autoconf.php is evaluated the first time is created

webian:
Resolves: #568


dmitryd:
Solved the other way.

PR #571, opened 2017-11-20
[BUGFIX] prevent cache entries for pages that are not configured

schugabe:
check if rootPageId is found on the root line otherwise do nothing

use emptyUrlReturnValue only if the page is a configured rootPageId

### DIFF
--- a/Classes/Encoder/UrlEncoder.php
+++ b/Classes/Encoder/UrlEncoder.php
@@ -589,12 +589,18 @@ class UrlEncoder extends EncodeDecoderBase {
 		/** @var \TYPO3\CMS\Core\Utility\RootlineUtility $rootLineUtility */
 		$rootLine = $rootLineUtility->get();
 
+		$foundRootPageId = false;
 		// Skip from the root of the tree to the first level of pages
 		while (count($rootLine) !== 0) {
 			$page = array_pop($rootLine);
 			if ($page['uid'] == $this->rootPageId) {
+				$foundRootPageId = true;
 				break;
 			}
+		}
+
+		if (!$foundRootPageId) {
+			throw new \RuntimeException('Configured rootPageId not in rootLine', 1506001361);
 		}
 
 		$languageExceptionUids = (string)$this->configuration->get('pagePath/languageExceptionUids');
@@ -991,7 +997,7 @@ class UrlEncoder extends EncodeDecoderBase {
 				$this->encodePathComponents();
 			}
 			catch (\Exception $exception) {
-				if ($exception->getCode() === 1343589451) {
+				if ($exception->getCode() === 1343589451 || $exception->getCode() === 1506001361) {
 					// Rootline failure: "Could not fetch page data for uid X"
 					// Reset and quit. See https://github.com/dmitryd/typo3-realurl/issues/200
 					$this->encodedUrl = $this->urlToEncode;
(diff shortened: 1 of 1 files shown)


liayn:
@dmitryd Ping. Would be nice to have that in the next version finally. Thanks.

dmitryd:
I am thinking on it. I do not see any obvious issues but something tells me there will be side effects.

liayn:
@dmitryd Not sure what you mean, because the title says it all. No matter how you manage to f*** up the configuration to get to this point, something like this must never be put into DB tables.

liayn:
@dmitryd Any reason you re-do the work on your own with 27dfbc236d6d3cc7f99cea183e618aea1e9f269d instead of getting this one in??

dmitryd:
@liayn 

> Any reason you re-do the work on your own with 27dfbc2 instead of getting this one in??

Because you are only preventing it but my code tries to recover. Better solution.

dmitryd:
@liayn 

200€ and I will answer all your questions. Don't want to pay? Than you got everything what you can get for free already.

liayn:
200€ and I will fix bug in TYPO3 core. Fair game.

dmitryd:
I agree. Ask T3A to pay you and see what they tell you... It is everybody's choice. I am supporting realurl for free for the past 12 years. I set hundreds of hours rewriting the most popular & most used TYPO3 extension. Did I get anything for doing that? Nope. So I am limiting my support to paying clients and do the rest only when I am in a very good mood.

Fair?

liayn:
Hey... lovely discussion... you know that I was one of the most active contributors for Core quite a while? Hey, I was stupid enough to do it for free. Say thanks for having a Core your extension can work with.

It is totally superfluous to discuss this. In an ecosystem of opensource it is vital to have collaboration on a voluntary basis to some degree. You know my point and you know what you would expect from someone where you report (and fix) an issue.
Let's end the conversation here.

dmitryd:
> you know that I was one of the most active contributors for Core quite a while? Hey, I was stupid enough to do it for free. Say thanks for having a Core your extension can work with.

Do you know I was the most active contributor for the most successful TYPO3 version, 4.5? :) And there is no part of the core, where I did not contribute in the past for many years. I was a core team member from around 2005 till at least 2014 I think. So do we really have to measure who contributed to TYPO3 most?

> It is totally superfluous to discuss this. In an ecosystem of opensource it is vital to have collaboration on a voluntary basis to some degree. You know my point and you know what you would expect from someone where you report (and fix) an issue.

The thing is: nothing from YOUR code was used in solving this. What kind of attribution do you expect? If you just provided a solution but it was not used, why do you think you have a right to attribution?

schugabe:
No this is not correct: I found the bug and implemented the PR. You reimplemnted it with differnt variable names and expanded it in a second commit. Obviously I found a valid bug and implemented a solution you used to add additional features. 

"Normal" open source projcts are glad if somebody subbmits a issue with a PR. It makes no sense if you complain about having no time and at the same time reimplment code that you also have in a working PR. I don't understand why you would contribute/maintain open source projects if you don't like the process (and probably people in general) but hey whatever. Hopefully we don't need RealUrl any more in the near future and you don't have to bother with this stuff any more.

dmitryd:
@schugabe 

> I found the bug and implemented the PR. You reimplemnted it with differnt variable names and expanded it in a second commit. Obviously I found a valid bug and implemented a solution you used to add additional features.

I made the fix as a part of other work for a paying client without looking to github at all. There is nothing common between your and my code. Thanks for finding but that does not mean attribution.

There are a lot of people here, whose code is merged and attributed and "thanks" said to them. But this is not the case here.

> I don't understand why you would contribute/maintain open source projects if you don't like the process

Everybody's reasons are different and no one has right to judge others.

> Hopefully we don't need RealUrl any more in the near future and you don't have to bother with this stuff any more.

I hope too because I am tired supporting this extension for the past 12 years. Just like Kasper before, I should step out. I only support this ext because no one else will be able to do it. And it is still the most used ext in TER. So it is a kind of duty that I do not want or like but just do because I took it from Kasper those 12 years ago.

And remember, you get realurl for free to use in your commercial projects. So at least say "thanks" from time to time.

Good luck!

dmitryd:
For me the discussion is closed as well. Sorry, guys, but I have more important things to do than to discuss this.

PR #572, opened 2017-11-20
[BUGFIX] check if only L param with cHash is present

schugabe:


dmitryd:
This should not be necessary because `L` is excluded from the parameter list by `getRelevantParameters()`.

liayn:
#504 would be the explanation for having L not being filtered. I have no clue though how this happens, but not being the only one with this, makes me feel confident that there is something going on indeed.

PR #574, opened 2017-11-27
[TASK] Add public getter for rootpage ID

### LINKED ISSUE #566: Add getter for root page ID
The property `$rootPageId` defined in `EncodeDecoderBase` and used by `UrlDecoder` is protected and thus not accessible from the outside. However, this can be useful for decode hooks.

Would you accept a PR which adds a public getter for this property?

mbrodala:
This can be used in decode hooks.

Fixes #566


dmitryd:
Thank you!

PR #575, opened 2017-12-04
Use default language if wrong one is provided and reduce log size

emmguyot:
To fix issue #549, when a bad language parameter is given, the default language is used.

A *short* error is logged to help in writing redirect action for fake URLs : This log is done only once per request.


pniederlag:
Would be nice if this PR would be accepted

pniederlag (changes requested):
Applying and testing the patch. It seems to do fine! :-> Thx

dmitryd:
Please, fix your site instead! It is really your job to provide proper urls to realurl, not realurl's job to recover for your implementation mistakes.

I already explained many times how to do it. I also have information on the wiki about it.

pniederlag:
> please, fix your site instead.

As we have no control over the L parameter that users might send to the TYPO3-system this can't be done without overly complex work. 

It is just bad practice to use debug_backtrace() in production context.

Nonetheless, thx for your engagement for realurl, As I understand you are not willing to change this, I must accept your decision while I still think it is wrong.

There are some workarounds plus a patch, which works nicely.

emmguyot:
Dmitry,

I agree with you that the site has to be fixed first.
I've checked again the wiki and I can't really find an article about this kind of problem. What are you thinking about for this kind of fix for the site ? Do you mean a configuration to redirect or handle such a "bad request" or a configuration for the log to prevent it from growing ?

In my case, the problem occurs because of several Typo3 upgrades and, even if now my site hasn't any problem when you crawl it, search engines has memorized some URL with problem with the L parameter. And the log grows each time they crawl these URLs. 
That's why my workaround was accurate for my problem, and probably for others too.

Please give us details on how to fix this kind of problems.

dmitryd:
@pniederlag

> It is just bad practice to use debug_backtrace() in production context.

It was requested by people, who could not get the information about the error in any other way. So this got integrated to help other people too. You can always send this output to nowhere with logging configuration.

artur2000:
@pniederlag 

Hi Dmitry, I understand the motivation behind adding the debug_backtrace() but this thing can actually generate really HUGE log files. As in a recent case from us, 3MB per log entry and 10GB per day.
Some other systems like e.g. magento also add debug_backtrace() to the log, but they actually truncate it to be just a few lines, so every DEV will get a start to debug it further, there is really no need for more.

Does that seem like a good compromise proposal?

dmitryd:
@artur2000 

> Does that seem like a good compromise proposal?

Well, sometimes the cause can be 50 lines deeper (like if you generate a link using the menu). So with a short log the log will be useless. We do a lot of projects and never had the issue. We control what we pass to links, that's all. Validation is the key. If the site lacks validation of GET parameters, it is pretty bad...

dmitryd:
@artur2000

> How can I validate what the site user types into the URL?

You can read the docs at least :) `config.linkVars` and avoiding `addQueryString` will make sure you never have wrong parameters in your urls.

Btw, there should be much less logging in 2.4.0.

dmitryd:
@artur2000 

Realurl cannot prevent the user from typing anything in the browser. However *you* can prevent bad `L` from coming to realurl. This is what *you* need to do.

PR #576, opened 2017-12-08
[FEATURE] Alias per page

mhuber84:
With this commit the page id is stored with each alias. So an alias can be
used more than once as it is identified by the page on which the record is shown.
With this commit the behaviour of aliases is similar to the beaviour of pages:
The path segment is unique inside of the parent page.
While decoding preVars we have no page id, so we have to ignore it.


dmitryd:
Not good, for example, if same news has to be shown on different pages.

I do not see any value in this change. Also it breaks existing behavior.

There has to be really good reasons to accept this.

PR #583, opened 2017-12-14
Enable 404 Error Handling for Userfunc

jmtpix:
Because of some old content in the Google Index i tried to get a 404 Error for no longer existing subpages configured via userfunc in the fixedPostVars.
So I update the UrlDecoder Class and added the following to the realurl configuration:
GETvar' => 'myVar',
'enable404forInvalidAlias' => 1,
'userFunc' => ....

If there was already an existing way to do this, I'm sorry I wasn't able to find it. :)


dmitryd:
Why can't userFunc throw error itself?

PR #586, opened 2017-12-22
[FEATURE] Add post-processing hook for convertToSafeString method

typecat:
Sometimes string conversion needs to be processed differently (e.g. for suomi URLs). This commit adds a hook to enabling to do so.


dmitryd:
Thank you!

PR #595, opened 2018-01-24
Give also the IP address

franzholz:
It is necessary to be able to track Dos attackers who misuse the realurl feature of writing continuously into the PHP error_log file.


DanielRuf:
This will be forbidden with the new GDPR law.

And the IP won't help you in most cases.

DanielRuf:
Also `HTTP_CLIENT_IP` is a value that is set by clients so you will get spoofed IP addresses in most cases (any HTTP_ header is set by the client).

dmitryd:
Rejecting due to GDPR

Issue #577, opened 2017-12-10
Shortcut pages redirected to decoded url


webian:
After commit https://github.com/dmitryd/typo3-realurl/commit/54c5233041f7ba272ac61a6c2c2f08950294c0c8 (fixing https://github.com/dmitryd/typo3-realurl/issues/548 & https://github.com/dmitryd/typo3-realurl/issues/570) when I browse a shortcut page, it redirects to the target but using the decoded url.

By instance, we have a page with title "shortcut" that has as target a page with title "target" and uid=123.
When I browse: `http://foo.bar/shortcut/` it should redirects to `http://foo.bar/target/` but instead it redirects to `http://foo.bar/index.php?id=123`

After a quick check, new method `getAbsRefPrefix()` prepends a slash to index.php while, before the commit this didn't happen.

dmitryd:
Can't reproduce this.

JNellissen-BM:
We have a similiar problems with real url and TYPO3 shortcut pages.
In our case it seems, there is a problem between realurl and the server config.
On our dev Systems everything is fine and the shortcut url will be  generated like expected, but on the client server every shortcut redirects to the standard TYPO3 url (index.php?id=XXX)
I can't figure out where are the differents (I have no direct access to the client server) and everything looks nearly identical, but I guess, some information from the request should be wrong, so the domain could not be determined correctly... or something else ;-)
I am sorry, I can't give you more information beside that I can confirm this problem,
or like we say in german "Irgendwo ist da der Wurm drin" ;-)

Kind regards
Jeffrey

fgerards:
its reproducable here to: after update to 2.3.1, shortcut target-url is not realurl-encoded anymore.
Instead, it redirects to index.php?id=<id>. This is a major bug and would cause google rankings to drop.
plz fix, thank you !

mbrodala:
I can also confirm the issue:

- 2.3.0: `/shortcut/` redirects to `/target/`
- 2.3.1: `/shortcut/` redirects to `/index.php?id=N`

Downgrading to 2.3.0 restores the correct behavior.

christian-hellmuth:
I can also confirm the described misbehavior. Not going into further detail as it's already been described correctly before.

automatrix:
I can also confirm this - I also had to downgrade to v. 2.3.0 to have it work again.

rafu1987:
I can also confirm this and a downgrade to 2.3.0 helped.

corth:
[...]

PR #598, opened 2018-02-08
[BUGFIX] shortcuts need to redirect to encoded URLs

AndreasA:
Resolves: #577


dmitryd:
Thank you very much :)

PR #600, opened 2018-02-12
[BUGFIX] Use makeInstance instead of $GLOBALS for CharsetConverter

rrrapha:
This is needed for compatibility with TYPO3 v9.


dmitryd:
FYI: realurl is not going to work with TYPO3 v9.

dmitryd:
Thanks!

PR #602, opened 2018-02-15
[BUGFIX] Use require for realurl_autoconf.php inclusion

rrrapha:
Resolves: #568


dmitryd:
Solved the other way.

Issue #606, opened 2018-03-07
$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT'] is now null


DanielRuf:
With the update ~2 hours ago `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT']` is not set by default now.

I guess this is a regression.

dmitryd:
Most likely you use `$TYPO3_CONF_VARS` instead of `$GLOBALS['TYPO3_CONF_VARS']` in your config.

DanielRuf:
> With the update ~2 hours ago $GLOBALS['TYPO3_CONF_VARS']



> Most likely you use $TYPO3_CONF_VARS instead of $GLOBALS['TYPO3_CONF_VARS'] in your config.

This is not the case.

DanielRuf:
Just take a look at the linked PR and the other comments which reference and comments at https://github.com/dmitryd/typo3-realurl/pull/517

dmitryd:
Sure, I did look to your PR:
```
$TYPO3_CONF_VARS['EXTCONF']['realurl'] = [
```

Not the case?..

DanielRuf:
I guess you miss the point.

You use `require_once` but forgot to define the globals.
See https://github.com/dmitryd/typo3-realurl/pull/517#issuecomment-328125980

https://github.com/dmitryd/typo3-realurl/commit/ab6a5cf5c5fed1bf804ed1119a469ed72406c0c6

DanielRuf:
This
`global $T3_SERVICES, $T3_VAR, $TYPO3_CONF_VARS;`
is clearly missing and so the current version breaks setups.

dmitryd:
Try using `$TYPO3_CONF_VARS` in `ext_localconf.php`. It worked before but now it does not. Didn't it break setups than? :) Always use `$GLOBALS`. It would be faster if you just go and fix it on your site than write comments here :)

DanielRuf:
> than write comments here :)

That's why I've provided a PR.
You forgot the `global` definitions. About 6 months ago you wrote that `require_once` is not enough and would be a regression and break setups.

3 months later you change it to `require_once`, the way that you said before is not the solution.

Your change will break other setups too. Don't you want to add the `global` definitions then or revert the change and apply the proposed change?

Not sure why you are like this but your patch release breaks several setups. And it is definitely a bug in your code (missing `global` definitions).

[...]

PR #607, opened 2018-03-07
[BUGFIX] use requireOnce of the GeneralUtility to load the default config

DanielRuf:
#606 

The change to `require_once` in 2.3.2 breaks TYPO3 8 (and older) setups where `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT']` is accessed for example in `typo3conf/RealUrlConfiguration.php` and results in `null` and not the expected `Array`.

Before this it did work. 2.3.1 was still fine.

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -29,11 +29,11 @@ if (!function_exists('includeRealurlConfiguration')) {
 		$realurlConfigurationFile = trim($configuration['configFile']);
 		if ($realurlConfigurationFile && @file_exists(PATH_site . $realurlConfigurationFile)) {
 			/** @noinspection PhpIncludeInspection */
-			require_once(PATH_site . $realurlConfigurationFile);
+			\TYPO3\CMS\Core\Utility\GeneralUtility::requireOnce(PATH_site . $realurlConfigurationFile);
 		}
 		elseif ($configuration['enableAutoConf'] && file_exists(PATH_site . TX_REALURL_AUTOCONF_FILE)) {
 			/** @noinspection PhpIncludeInspection */
-			require_once(PATH_site . TX_REALURL_AUTOCONF_FILE);
+			\TYPO3\CMS\Core\Utility\GeneralUtility::requireOnce(PATH_site . TX_REALURL_AUTOCONF_FILE);
 		}
 
 		if (is_array($existingConfiguration)) {


DanielRuf:
Recommendation of a colleague of me: add a version check and only use the new code for 9 and newer.

Currently the extension has a version constraint < 9.0.0 so we should not yet change this line. Also this is much more than a SemVer patch ;-)

dmitryd:
`RealUrlConfiguration.php` is not the file I know about. Please, provide more information about the problem. I tested under 7.6 and 8.7, works fine here.

DanielRuf:
> RealUrlConfiguration.php

typo3conf/RealUrlConfiguration.php

DanielRuf:
But this is just an exemplary config file name.
In there `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT']` returned the default conf array in 2.3.1. With 2.3.2 it returns `null`.

The suggested change fixes it in TYPO3 8.

DanielRuf:
> is accessed for example in `RealUrlConfiguration.php`

"for example" ;-)

DanielRuf:
```
<?php
...

$TYPO3_CONF_VARS['EXTCONF']['realurl'] = [
    'encodeSpURL_postProc' => ['user_encodeSpURL_postProc'],
    'decodeSpURL_preProc' => ['user_decodeSpURL_preProc'],
    '_DEFAULT' => [
    // some default config here
    ],
];

// Domains and Subdomain configuration
$sites = [
    'sitename' => [
        'domains' => [
          
        ],
        'customConf' => [
        ],
    ],
];

foreach ($sites as $key => $site) {
    $configuration = $GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl']['_DEFAULT'];
    $configuration = array_replace_recursive($configuration, $site['customConf']);
    foreach ($site['domains'] as $domain) {
        $GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl'][$domain] = $configuration;
    }
    unset($configuration);
}
unset($sites);
```

This worked in 2.3.1 and before and is broken in 2.3.2 due to the changes with `require_once`.

AlexHartung:
@DanielRuf your fix worked for us as well. We faced the same issue with 2.3.2.

DanielRuf:
@AlexHartung thanks for the feedback.

So far it was breaking deployments for TYPO3 systems in development and will also affect production systems in the long term when they install the update which can be a real blocker.

halliballi123:
Since GeneralUtility::requireOnce is deprecated, I think the problem should be solved by prepending $GLOBALS in the config array. Then you can use require_once and everthing just works fine.

change 
$TYPO3_CONF_VARS['FE']['addRootLineFields'].= ',tx_realurl_pathsegment';
$TYPO3_CONF_VARS['EXTCONF']['realurl'] = array(
...
...
);

$GLOBALS['TYPO3_CONF_VARS']['FE']['addRootLineFields'].= ',tx_realurl_pathsegment';
$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['realurl'] = array(
...
...
);

DanielRuf:
> Since GeneralUtility::requireOnce is deprecated, I think the problem should be solved by prepending $GLOBALS in the config array. Then you can use require_once and everthing just works fine.

This is not the right solution.
You can and will not change the configs in all cases.
Adding `global` keyword is the solution.

Also the deprecation does not count here as realurl is bound to `typo3cms <= 9`

DanielRuf:
Also check https://github.com/dmitryd/typo3-realurl/issues/606 and the links in my comments, especially https://github.com/dmitryd/typo3-realurl/issues/606#issuecomment-371855973

DanielRuf:
An alternative and better solution is to add the missing `global` definements.

zeno-codekunst:
@DanielRuf Do you already have a plan when your make your next tagged release that is including this fix? And when will it be?

DanielRuf:
> @DanielRuf Do you already have a plan when your make your next tagged release that is including this fix? And when will it be?

I'm just a contributor and not the maintainer who is @dmitryd, only he can tell.

PR #614, opened 2018-03-30
UrlDecoder->convertAliasToId() correct lookup

0bserver:
Use LIKE '%my%%path%%segment%' pattern for query that will yield the DB row
needed much more likely than a query with the unaltered path segment in the
WHERE clause.


dmitryd:
What??? Can you write that so that I can understand it?

dmitryd:
Too many hacks. Sorry, won't include it.

PR #617, opened 2018-04-05
[TASK] Make URLs case sensitive

### LINKED ISSUE #482: Case insensitive URLs
Hello,

For some reason all urls are case insensitive
http://www.site.de/ErroR-404/
http://www.site.de/eRror-404/
http://www.site.de/error-404/

Can you please suggest me how to disable it?
Thanks

stucki:
As explained in #482 RealUrl is case sensitive for non-cached URLs, but it accepts different case spelling for cached results.

Scenario:
1. Open http://mysite/HOME/ => it fails
2. Open http://mysite/home/ => page is loaded, cache is filled
3. Open http://mysite/HOME/ again => page is loaded

This is problematic because it leads to duplicate URLs for the same content.

Solution: When a URL is fetched from the cache which does not match the case spelling, treat it as expired and redirect to the correct URL.

Closes #482

### DIFF
--- a/Classes/Decoder/UrlDecoder.php
+++ b/Classes/Decoder/UrlDecoder.php
@@ -542,7 +542,7 @@ class UrlDecoder extends EncodeDecoderBase implements SingletonInterface {
 			$remainingPathSegments = $savedRemainingPathSegments;
 		}
 		if ($result && $this->expiredPath) {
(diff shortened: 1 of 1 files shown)


stucki:
The problem remains but since the basis for this was changed in RealUrl 3.2.2, I created a new PR for it. See #618

PR #620, opened 2018-04-18
Allow PHP 7.2

markuspoerschke:
I cannot see any reason why realurl is not compatible with PHP 7.2. We are using it with PHP 7.2 and we could not see any problems so far.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     },
     "require": {
         "typo3/cms-core": ">=6.2.0,<9.0.0",
-        "php": ">=5.4.0,<7.2.0",
+        "php": ">=5.4.0,<7.3.0",
         "ext-mbstring": "*"
     },
     "require-dev": {
--- a/ext_emconf.php
+++ b/ext_emconf.php
@@ -26,7 +26,7 @@ $EM_CONF[$_EXTKEY] = array (
     'depends' => 
     array (
       'typo3' => '6.2.0-8.9.999',
-      'php' => '5.4.0-7.1.999',
+      'php' => '5.4.0-7.2.999',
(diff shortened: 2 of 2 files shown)


LeoniePhiline:
Fantastic, thanks!

dmitryd:
> I cannot see any reason why realurl is not compatible with PHP 7.2.

Neither do I. I just do not have the setup to test it.

markuspoerschke:
@dmitryd Will you merge this pull request?

tantegerda1:
Shouldn't the same go into [ext_emconf.php](https://github.com/dmitryd/typo3-realurl/blob/development/ext_emconf.php#L29)?

markuspoerschke:
@tantegerda1 Good catch! Thanks! I changed it.

From my side this can be merged. :)

iamandrewluca:
`7.2.4` is already out :|
```
dmitryd/typo3-realurl 2.3.2 requires php >=5.4.0,<7.2.0 -> your PHP version (7.2.4) does not satisfy that requirement.
```

gilbertsoft:
can confirm works here too for multiple installations

JesperBom:
No errors here aswel in multiple installations

Nemo64:
This should be merged asap since the last version compatible with php `7.2` is realurl `2.3.1` which has major problems with shortcuts. For the time being I have to use `"conflict": {"dmitryd/typo3-realurl": "2.3.1"}` to avoid those issues as they don't exist in `2.3.0`.

dmitryd:
You can now use `dev-development` in composer for realurl if you urgently need 7.2 support.

PR #636, opened 2018-07-24
Add pre-processing hook in convertToSafeString

schliesser:
As we discussed in slack, I added the pre-processing hook in convertToSafeString


dmitryd:
Thank you!

Issue #601, opened 2018-02-15
Ugly URL's with complex postVarSets with all defined GETvars


tobiasklepp:
After updating from realkurl 1.x to 2.x I run into a big issue of changing behaviour, how URLs are generated. I read all the available documentation to find out, if my configuration is wrong, but I cant find an error. Looks like the behaviour changes and the result is in my opinion not good. I use TYPO3 8.7.10, PHP 7.0, MySQL 5.5 and realurl 2.3.1.

I have realurl configured in the init section with `'emptySegmentValue' => '-',`. I need this configuration, because otherwise urls didn't work correct, if they are inside of an email or will be blocked by some firewalls, if multiple slashes following with no values between. For example MS Outlook automatically changes url like http://www.domain.com/page/my_key///2/ to http://www.domain.com/page/my_key/2/. So the plugin receives to value 2 not for skip but for show.

For the postVarSets I use the following configuration:

```
'postVarSets' => [
    '_DEFAULT' => [
        'my_key' => [
            [
                'GETvar' => 'tx_myext_pi1[show]',
            ],
            [
                'GETvar' => 'tx_myext_pi1[episode]',
            ],
            [
                'GETvar' => 'tx_myext_pi1[skip]',
            ],
            [
                'GETvar' => 'tx_myext_pi1[media]',
            ],
            [
                'GETvar' => 'tx_myext_pi1[sort]',
            ],
        ],
    ],
],
```
The URL looks now like this.
TYPO3: `index.php?id=1&tx_myext_pi1[show]=1`
Realurl 2.x: `http://www.domain.com/page/my_key/1/-/-/-/-/`
Realurl 1.x: `http://www.domain.com/page/my_key/1/`

If I combine it also with a output type, ge get the following url.
TYPO3: `index.php?id=1&tx_myext_pi1[show]=1&type=444`
Realurl 2.x (emptySegmentValue: -): `http://www.domain.com/page/my_key/1/-/-/-/-/rss.xml`
Realurl 2.x (emptySegmentValue: not set): `http://www.domain.com/page/my_key/1/////rss.xml`
Realurl 1.x: `http://www.domain.com/page/my_key/1/rss.xml`

The url `http://www.domain.com/page/my_key/1/////rss.xml` can not handled by Apache and return an 500 Internal Server Error.

**Is there a missing or wrong configuration on my side or is this the wanted behaviour?**

dmitryd:
[...]

PR #637, opened 2018-08-02
Adds support for remove of trailing empty segments.

tobiasklepp:
Solves issue #601


davdenic:
hi @tobiasklepp I'm trying your solution but to me it does not work :(

tobiasklepp:
> 
> 
> hi @tobiasklepp I'm trying your solution but to me it does not work :(

Hi @davdenic, if you like you can use my fork: https://github.com/hope-media/typo3-realurl

dmitryd:
Not acceptable: it changes vendor name space in composer and extension version.

dmitryd:
The fix is incorrect because it removes empty segments from a single postVar, thus breaking the case when there are multiple postVars.

Proper solution for your case would be one of:

* Using `init/emptySegmentValue` configuration option
* Using separate postVars

Since this cannot eb integrated, I close this PR.

PR #650, opened 2018-10-21
Ignore new Facebook Urlparameter fbclid

### LINKED ISSUE #649: Ignore new Facebook Url Parameter - fbclid

WorksDev:
- Fixes #649

### DIFF
--- a/Classes/Configuration/ConfigurationReader.php
+++ b/Classes/Configuration/ConfigurationReader.php
@@ -85,7 +85,7 @@ class ConfigurationReader {
 	 */
 	protected $defaultValues = array(
 		'cache/banUrlsRegExp' => '/tx_solr|tx_indexedsearch|tx_kesearch|(?:^|\?|&)q=/',
-		'cache/ignoredGetParametersRegExp' => '/^(?:gclid|utm_(?:source|medium|campaign|term|content)|pk_campaign|pk_kwd|TSFE_ADMIN_PANEL.*)$/',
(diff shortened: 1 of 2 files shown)


jonaseberledmind:
Hi @dmitryd,

don't you think it would be better to use the already configured TYPO3 setting for that? So that realurl does not have to be patched (or the site config changed) if another param shall be excluded.




```
...
                'cache' => [
                    'ignoredGetParametersRegExp' => $this->ignoreCHashExcludedParametersRegex(),
                ],
...

    protected function ignoreCHashExcludedParametersRegex()
    {
        return '/^(?:'
            . implode(
                '|',
                array_map(
                    function ($e) {
                        return preg_quote($e, '/');
                    },
                    $GLOBALS['TYPO3_CONF_VARS']['FE']['cacheHash']['excludedParameters']
                )
            )
            . ')$/';
    }
```

in my realurl-hook

jonaseberle:
Please try this: Add a parameter in the Install Tool in ignoreCHashExclude... 
When realurl is enabled, it throws a 404 when the parameter is added to an URL.
When disabled, no 404.

Of course those 2 parameters are linked.

dmitryd:
jonaseberle, I am sure you know realurl better than I do ;)

jonaseberle:
I was just making a suggestion. One addition: exclude L from the list above.

dmitryd:
`L` is already ignored by `CacheHashCalculator` class. There is no need to add it there.

PR #655, opened 2018-11-28
cleanup $this->originalUrlParameters['L']

SvenDeKa:
now using $this->sysLanguageUid instead -looked forgotten, as was is used 
in createPathComponentThroughOverride
but not in createPathComponentUsingRootline()

using originalUrlParameters['L'] blocked me from using an alternative Language-Param with own behaviour

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
-    "name": "dmitryd/typo3-realurl",
+    "name": "svendeka/typo3-realurl",
     "type": "typo3-cms-extension",
-    "description": "Speaking URLs for TYPO3",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,7 +1,7 @@
 {
-    "name": "dmitryd/typo3-realurl",
+    "name": "svendeka/typo3-realurl",
```
mmunz: I'm relatively sure this shouldn't be changed ;)

SvenDeKa: yeah, your right. testing purposes. will be changed back. thx!

PR #656, opened 2018-12-17
Prevent iteration over null value

sascha-egerer:


dmitryd:
It is not enough. Please, describe when & how it happens.

sascha-egerer:
I don't care anymore so I'll close this pull request.

Issue #669, opened 2019-08-07
PHP 7.3 compatibility


instruccionesaldorso:
Can the latest version of the extension be checked against PHP 7.3 compatibility?

https://github.com/dmitryd/typo3-realurl/blob/development/ext_emconf.php#L29
https://github.com/dmitryd/typo3-realurl/blob/development/composer.json#L28

LeoniePhiline:
@dmitryd We are using the extension for quite a while on PHP 7.3.
Please release a new version with adjusted restraints for 7.3 on packagist, so everyone can let go of the obnoxious `--ignore-platform-reqs`.

Thank you!

instruccionesaldorso:
Any update on this issue? The only extension with no PHP 7.3 restraint we have in our systems is this one.

PowerKiKi:
This is a duplicate of #657 and should be closed.

dmitryd:
There was a report in TYPO3 slack that somebody had problems with php 7.3 and regexp used in realurl.

dmitryd:
Merged the PR.

Issue #657, opened 2019-01-30
Set compatibility to PHP 7.3.x


MartinH0:
I'm using this extension under PHP 7.3.1 at Typo3 8.7.x and have never had a problem! Everything is working fine as it should :)
Pls change compatibility to PHP 7.3 so also People who run this extension with PHP 7.3 get the Updates, if not Updates are not shown in extensionmanager as requirements are not meet

dmitryd:
There is another ticket where people have warnings with 7.3. If you do not have the problem, it does not mean nobody else will. This has to be tested before the version can be raised.

M4rt1n17:
@dmitryd thanks. Then I would please you to see this as "it's working for me".

LeoniePhiline:
@dmitryd Searching for "7.3" or "Warning" in the open issues yielded no result. Is the ticket you mentioned closed by now?

PR #671, opened 2019-09-16
[TASK] Mark as compatible with PHP 7.3

smichaelsen:
According to PHPStan and to our tests realurl has no compatiblity issues with PHP 7.3


PowerKiKi:
This will solve #657 and #669.

